PR #9, opened 2022-07-07
make mace compatible with pytorch lightning

felixmusil:
Hello,

I am using pytorch_lightning to train my models and it requires only a few aesthetic changes to the code base, namely:

+ switch activations from functional to module, e.g. `torch.nn.functional.silu -> torch.nn.SiLU`

+ remove device arg from `one_hot_encoding` since it has to be consistent with the input argument

+ remove dtype from MACE's attribute (type should be managed through `.to()` and it's not used and can't be easily serialized) 

+ put the U_matrices in the buffers so that they are moved to the right device with
the model

+ remove device argument from MACE model

The tests and linter are passing.

The content of this PR is provided under [MIT license](https://opensource.org/licenses/MIT)"


felixmusil:
Is there any more to add ?

PR #14, opened 2022-09-05
Dipole MACE model

davkovacs:
This pull request implements a MACE model that can predict l=1 equivariant outputs, molecular dipoles as the most trivial example.

### DIFF
--- a/mace/calculators/__init__.py
+++ b/mace/calculators/__init__.py
@@ -1,3 +1,3 @@
-from .mace import DipoleMACECalculator, MACECalculator
-
-__all__ = ["MACECalculator", "DipoleMACECalculator"]
+from .mace import DipoleMACECalculator, MACECalculator, EnergyDipoleMACECalculator
+  
+__all__ = ["MACECalculator", "DipoleMACECalculator","EnergyDipoleMACECalculator"]
--- a/mace/calculators/__init__.py
+++ b/mace/calculators/__init__.py
@@ -1,3 +1,3 @@
-from .mace import MACECalculator, DipoleMACECalculator
+from .mace import DipoleMACECalculator, MACECalculator
 
 __all__ = ["MACECalculator", "DipoleMACECalculator"]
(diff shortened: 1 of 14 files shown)


davkovacs:
An example input script for fitting a dipole MACE model:
```sh
python ./mace/scripts/run_train.py \
    --name="MACE_dipole" \
    --train_file="train.xyz" \
    --valid_fraction=0.05 \
    --test_file="test.xyz" \
    --charges_key="Qs" \
    --dipole_key="dipole" \
    --model="AtomicDipolesMACE" \
    --num_interactions=2 \
    --max_ell=3 \
    --hidden_irreps='32x0e + 32x1o' \
    --num_cutoff_basis=5 \
    --correlation=3 \
    --r_max=4.5 \
    --batch_size=16 \
    --valid_batch_size=64 \
    --eval_interval=5 \
    --max_num_epochs=1500 \
    --ema \
    --ema_decay=0.99 \
    --amsgrad \
    --loss="dipole" \
    --error_table='DipoleRMSE' \
    --default_dtype="float32"\
    --device=cuda \
    --seed=12345 \
```

davkovacs:
As a word of caution, a system that has an overall charge the predicted dipole moment will have a component that depends on the choice of the origin. This is dealt with by requiring the specification of atomic charges, that sum to the correct total charge. If atomic charges are not specified they will be set to 0 by default, so that the predicted dipole will not have a contribution from them (it will be purely a sum of atom-centred dipole vectors). This will work well for neutral systems.

davkovacs:
Example for using the ASE calculator to compute dipoles:
```py
import ase
from ase.io import read
from mace.calculators import DipoleMACECalculator

at = read("test.xyz", "0")
calc = DipoleMACECalculator(model_path="checkpoints/MACE_dipole_run-12345.model", device="cpu", charges_key="Qs")
at.set_calculator(calc)
print(at.info["dipole"])
print( at.get_dipole_moment())
```

Output:
```sh
>>> array([22.64881749, 27.27163923, 24.56755236])
>>> array([[22.758026, 27.163475, 24.666084]], dtype=float32)
```

ilyes319:
It looks great to me. I will just need to translate these models to torchscript also.

PR #22, opened 2022-09-22
Per config energy/forces/stress/virials weight

### LINKED ISSUE #13: Error due to missing virials on some configurations
Using the `stress_develop` with the following input script and the attached `train.xyz` and `test.xyz` (renamed to add `.txt` extensions) leads to an error.

[test.xyz.txt](https://github.com/ACEsuit/mace/files/9488752/test.xyz.txt)
[train.xyz.txt](https://github.com/ACEsuit/mace/files/9488753/train.xyz.txt)

```bash
python -i ../mace/scripts/run_train.py \
    --name="MACE_model" \
    --train_file="train.xyz" \
    --valid_fraction=0.05 \
    --test_file="test.xyz" \
    --config_type_weights='{"Default":1.0}' \
    --model="MACE" \
    --hidden_irreps='128x0e + 128x1o' \
    --r_max=5.0 \
    --batch_size=10 \
    --max_num_epochs=1500 \
    --swa \
    --start_swa=1200 \
    --ema \
    --ema_decay=0.99 \
    --amsgrad \
    --device=cuda \
    --forces_key=dft_force \
    --energy_key=dft_energy \
    --virials_key=dft_virial \
    --loss=virials
```

Traceback is as follows, with interactive debugging that it its `ref["virals"]` which is `None`.

```
2022-09-05 10:51:12.044 INFO: Started training
  File "../mace/scripts/run_train.py", line 405, in <module>
    main()
  File "../mace/scripts/run_train.py", line 346, in main
    tools.train(
  File "/gpfs/home/e/essswb/mace-venv/lib/python3.8/site-packages/mace/tools/train.py", line 69, in train
    _, opt_metrics = take_step(
  File "/gpfs/home/e/essswb/mace-venv/lib/python3.8/site-packages/mace/tools/train.py", line 206, in take_step
    loss = loss_fn(pred=output, ref=batch)
  File "/sulis/easybuild/software/PyTorch/1.9.0-fosscuda-2020b/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1051, in _call_impl
    return forward_call(*input, **kwargs)
  File "/gpfs/home/e/essswb/mace-venv/lib/python3.8/site-packages/mace/modules/loss.py", line 172, in forward
    + self.virials_weight * weighted_mean_squared_virials(ref, pred)
  File "/gpfs/home/e/essswb/mace-venv/lib/python3.8/site-packages/mace/modules/loss.py", line 41, in weighted_mean_squared_virials
    configs_weight * torch.square((ref["virials"] - pred["virials"]) / num_atoms)
  File "/sulis/easybuild/software/PyTorch/1.9.0-fosscuda-2020b/lib/python3.8/site-packages/torch/_tensor.py", line 544, in __rsub__
    return _C._VariableFunctions.rsub(self, other)
TypeError: rsub() received an invalid combination of arguments - got (Tensor, NoneType), but expected one of:
 * (Tensor input, Tensor other, *, Number alpha)
 * (Tensor input, Number other, Number alpha)
```

bernstei:
Much deeper in the torch code, so not nearly as trivial as my last suggestions.  Does it seem like it might be right? I know very little about torch.

closes #13, now that it creates property but sets it weight to 0 automatically


bernstei:
Definitely broken. For example, `tests/test_modules.py::TestLoss::test_weighted_loss` complains 
```
        configs_weight = ref.weight  # [n_graphs, ]
>       configs_energy_weight = ref.energy_weight  # [n_graphs, ]
E       AttributeError: 'Batch' object has no attribute 'energy_weight'
```
which is all fine and well, I guess I missed someplace, except I don't even understand where the regular `weight` attribute, used in the line above, is added to the `Batch` object. Every place I saw `weight` I added and analogous construction for `energy_weight` (and so on for F/S/V).

ilyes319:
I have added small fixes, it should be working now

bernstei:
Thanks - I have no idea how my obviously wrong `AtomicData.__init__` made it through. Now we can handle missing data by setting it to anything, and setting the weight to 0. Is that worth adding to this PR?

bernstei:
> Yes I think it would be great. You mean in case no stress or virials (or else) are found, we just set the corresponding tensor to zeros for example?

Exactly. Just the the quantity's value to anything (might as well be 0), and set the corresponding quantity-specific weight to 0 so it doesn't matter.

bernstei:
What do you think of the missing data implementation I just pushed?  Of course you could do something a lot more elaborate by sorting the configs by which data they have, and making sure each `Batch` is homogeneous, but I don't think that's worth the effort.

bernstei:
I'm working on a pytest of `run_train.py`. It's almost done, but I found a bug in setting `stress_key` and `virial_key`, which I'll include in this PR unless someone wants to cherry-pick it separately.  Also, is it true that `MACECalculator` cannot compute stresses yet?

bernstei:
pytest of `run_train.py` created, but it only checks that a model file was created, and that it can be used to instantiate a `MACECalculator` - nothing about the correctness of the predictions of the model yet.

ilyes319:
Thanks Noam, great idea for the run_train.py test. I think the calculator can not do stress yet, but it is a trivial addition that should be made.

ilyes319:
I think it is a good fix for now, I will test it and merge afterwards if you think it is enough.

ilyes319:
All the test are now passing on my setup. Could you check it works for you also?

bernstei:
I merged from develop to resolve the conflict, and all tests pass for me. I'm ready to merge, and then we can see what's going on with the following 2 PRs.  If you're still having issues with the pytest `run_train.py` path, I'd be happy to take a look if you describe the specifics here.

ilyes319:
Thanks for that. I still have the issue of mace not being correctly added to my path during the `run_train.py` pytest. Adding this line fixed it for me:  `run_env["PYTHONPATH"] = str(Path(__file__).parent.parent)`

bernstei:
Anyway, I'm happy to merge the way it is right now, with the directory relative to `__file__` explicitly added (as long as it's added to `sys.path`, since I need the subprocess to also have access to some other dirs in that list), but I was just surprised that it's not sufficient for you.

bernstei:
Actually, the current code isn't right - if it is inserted manually it should be prepended, not appended, so the testing version overrides any other installed versins.

bernstei:
That will _definitely_ break my system, because I need other things in the original `PYTHONPATH`, and your catch clause overrides it.  Can you try the version that I'm about to push, and print the output line that contains `"DEBUG"` if it's failing?

ilyes319:
Re on how I am running the tests : I am running them from a notebook, which might be a weird setting, I admit. I will try from terminal.

bernstei:
It'd be better if they worked both ways. How do you run `pytest` in a notebook?

bernstei:
It'd be better if they worked both ways. How do you run `pytest` in a notebook?

LarsSchaaf:
I get the following error when training on my sets:

```python
2022-10-17 10:51:50.020 INFO: Started training
  File "/home/lls34/rds/hpc-work/programms/mace-pr/scripts/run_train.py", line 490, in <module>
    main()
  File "/home/lls34/rds/hpc-work/programms/mace-pr/scripts/run_train.py", line 429, in main
    tools.train(
  File "/rds/user/lls34/hpc-work/programms/mace-pr/mace/tools/train.py", line 87, in train
    valid_loss, eval_metrics = evaluate(
  File "/rds/user/lls34/hpc-work/programms/mace-pr/mace/tools/train.py", line 289, in evaluate
    delta_stress_list.append(batch.stress - output["stress"])
TypeError: unsupported operand type(s) for -: 'Tensor' and 'NoneType'

```

I assume this is because missing data is filled in automatically since [bda6c53](https://github.com/ACEsuit/mace/pull/22/commits/bda6c53e544032776f636d71941a758e40812234), so `batch.stress is not None` is never None. Changing the if statements in `train.py: evaluate` 

from: 
```python
if "stress" in output.keys() and batch.stress is not None:
```
to 

```python
if "stress" in output.keys() and output["stress"] is not None and batch.stress is not None:
```

or
```python
if "stress" in output.keys() and output["stress"] is not None:
```

does the trick for me

bernstei:
Having looked a bit more at the code, my real question is how come `"stress"` is in `output.keys()`, but it has value `None`?  We fill in fake stress in the fitting _data_ (which is presumably what's in `batch`), but why does that affect the _output_?

bernstei:
@LarsSchaaf it might be worth it to figure out exactly what combination of data and training flags are leading to this error, so they can be added to the tests.

bernstei:
OK - I'm starting to see what's going on.  The normal model `ScaleShiftMACE` always seems to have `"stress"` in the output (line https://github.com/ACEsuit/mace/blob/0c0fbae7da8d60127adb9223c49539f06c919b7a/mace/modules/models.py#L318), as of `a2e85235 (Ilyes Batatia 2022-08-29 19:28:44 +0200 318)`.  Presumably that means that the output always contains a `stress` field, but it has value `None` if `compute_stress` isn't True.

The error @LarsSchaaf posted doesn't really seem to be specifically related to the filling in of stress - shouldn't it always have happened if you passed data with stress (so it's in `batch`) but also pass `compute_stress=False` to `tools.train()` (so there's no meaningful value in `output`)?  If that's true, I think the correct fix is to replace the
```
        if "virials" in output.keys() and batch.virials is not None:
```
with 
```
    if output.get("virials") is not None and batch.virials is not None:
```
which checks both for existence of the field and whether it contains a meaningful value.  If someone can confirm, I'd be happy to make those fixes to this PR.

bernstei:
My latest commit shows this error when I just take the missing data test and change the loss to `"weighted"`. I'm pretty sure the fix I suggest above is right, but I'd appreciate a confirmation from someone who thinks they understand the code design.

ilyes319:
Hey Noam, you are right, this is a correct fix. Thank you.

LarsSchaaf:
True that is a more elegant way. I quickly changed to

```
    if output.get("virials") is not None and batch.virials is not None:
```
 and it works for my test case. Note that I had no stresses at all in my training set as you described above. 
 
Just changing the code, I feel like these things should all be written in the form of a loop over possible outputs, no?

bernstei:
I'm testing the same fix right now, and will push if tests pass locally.  

The problem with writing it as a loop is that different quantities are stored in different ways, so I don't think it'd be any simpler.

ilyes319:
I am happy to merge now.

PR #24, opened 2022-09-30
ASE calculator stress

bernstei:
Add prediction of stress to ASE calculator

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -15,7 +15,7 @@ from mace.tools import torch_geometric, torch_tools, utils
 class MACECalculator(Calculator):
     """MACE ASE Calculator"""
 
-    implemented_properties = ["energy", "forces"]
+    implemented_properties = ["energy", "forces", "stress"]
 
     def __init__(
         self,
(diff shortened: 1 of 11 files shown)


ilyes319:
I think we should add some checks to catch the None?

bernstei:
which `None`?

bernstei:
I think I see what you mean `out["stress"]` is sometimes None.  When would that happen?

ilyes319:
There are flags called `compute_stress` that are inputed to the model and if they are False it outputs None. So we should link the properties required by the user to the flag. I don't really know how this is done in the calculator.

bernstei:
Where exactly are they supposed to be passed to the torch model at the `calculate` (prediction) stage? `out model(batch, compute_stress=True)` ?

ilyes319:
yes exactly.

bernstei:
OK - I think I know how to do that (`"stress" in properties`), but I'll double check and check something in, hopefully soon.

bernstei:
What's the cost of doing a separate predict for energy and another for forces (and stresses), vs. always (or at least by default) doing all 3 and caching them (since ASE has separate calls to retrieve each of those quantities)?

davkovacs:
I think we can compute all of them and cache the result if only energy is asked for. It is unlikely anyone will use the ASE calculator for some heavy work which only needs the energy.

bernstei:
Something's not working.  I'm calling
```
        out = self.model(batch, compute_force=True, compute_stress=True)
        print("BOB out", out)
```
and getting
```
BOB out {'energy': tensor([0.], grad_fn=<AddBackward0>), 'forces': None, 'virials': None, 'stress': None}
```

Do I need to do something special when fitting?

bernstei:
Also, it might be nice to make `force` and `forces` consistent everywhere.  I'm always guessing whether any given variable/flag is singular/plural.  And `virials` should probably always be singular, consistent with `energy` and `stress` (unless it's a separate virial associated with each atom).  Might also be nice to return individual atoms' energies, BTW.

bernstei:
Yes - when I pass `compute_stress` to the `model.batch` calls in `MACECalculator.calculate`, the _forces_ are not returned. Looks like a bug related to `compute_virial` vs `compute_stress` in the `if elseif` in `get_forces()`.  This would all be cleaner if internally you picked one or the other, and remove the one you're not using (except maybe at the top-most level).

PR #27, opened 2022-10-03
mace_fit executable

bernstei:
Move `run_train.py` to a directory under `mace/`, and add an entry point for it in `setup.cfg` so it's installed (under whatever pip install prefix) under `bin`


bernstei:
I'd like to make the final model export more versatile, too.  Is it true that if we save the cpu model it can always be loaded and then converted to cuda form, but if a cuda model is exported, it can only be loaded on a machine that has a GPU?

PR #32, opened 2022-10-20
Saving checkpoint before and after swa

davkovacs:


davkovacs:
This pull request implements the following:

- save the model with lowest validation loss both before and after swa
- evaluate both of the best models at the end of the training
- save both models as `.model` at the end of the training
- restart latest from swa checkpoint if present

davkovacs:
There are two more bugs to be fixed before merging: 

- when the second checkpoint (swa checkpoint) is loaded it throws error because of model and data not being on the same device ???
- The swa learning rate flag seems to be ignored.

PR #33, opened 2022-10-24
Change stress to Voigt notation

vue1999:
Changed the stress return format to align with the standard ase format by using only the 6 independent terms (Voigt notation).


davkovacs:
Looks good to me. @ilyes319 ?

PR #43, opened 2022-11-08
use 'aaargs' for dataclass like Argument Parsing

PythonFZ:
Hey MACE developers,

I've been playing around with your package recently and it produces some nice fits and very fast MLIPs for the systems I'm interested.
I'm often running experiments from within Python and found it a little bit challenging to do so.
With this PR I wanted to introduce a way to run a model either from within Python or via CLI.
It provides you with attribute autocomplete for the `args` class.

The changes involve:
- Addition of the very lightweight [aaargs](https://github.com/zincware/aaargs) package
- Introduction of the dataclass like `MaceArguments` class that holds all arguments
- move the actual `parse_args()` outside of the `main()` function. This allows to usage of `main()` from within Python

This PR does not affect anything outside the `run_train.py` / `arg_parser.py`.
Some examples exceeding the ones for the aaargs pacakge would be:

```python
from run_training import main
from mace.tools import MaceArguments

run_training(MaceArguments(name="Exp01", train_file="data.extxyz" ))
```

Regarding the ASL License you can use this PR in every way you'd like. I relinquish any authorship over the code in this PR.


ilyes319:
Hey @PythonFZ,

Very happy to read that you are using MACE for your projects and thanks for your PR. As we already have a PR #27 changing some aspect of the `run_train.py`, the merge might be a bit delay until we sort this one, but we will have a careful look.

bernstei:
@PythonFZ I see the advantage of moving the parsing of arguments outside of `__main__` and passing in the `args` namespace, but what's the advantage of `aaargs` over `argparse`?

PythonFZ:
> @PythonFZ I see the advantage of moving the parsing of arguments outside of `__main__` and passing in the `args` namespace, but what's the advantage of `aaargs` over `argparse` (asking because I'm working on another, unrelated package and considering cleaning up my parsing as well. We've also used https://click.palletsprojects.com/en/8.1.x/ before) ?

The main reason I thought something like `aaargs` could be great for you is, because it allows the user to interface with your code either through CLI or via a Python interface. I think it is primarly a cleaner way to write  CLI using argpase in the backend and support interfacing in a dataclass like fashion as described above by

```python
from run_training import main
from mace.tools import MaceArguments

run_training(MaceArguments(name="Exp01", train_file="data.extxyz" ))
```

You can:
- easily create instances of the `MaceArguments` class (compared to `parser.parse_args(["--train_file", "data.extxyz", ...])`)
- you have attribute autocompletion inside your IDE
- the available arguments and defaults are more clear.
- you can serialize from / to `yaml` or `json` files
- it is very simplistic and small compared to other CLI libraries
- do what `argparse` can in a cleaner and more easy to read way.

In comparison to `click` or `typer` the `aaargs` library doesn't do any of the CLI handling itself. It just wraps `argparse`.
In other words, you will struggle with nested CLI arguments and all the other nice stuff that these libraries implement.

So my personal suggestion is, that if you like the simplicity of `argparse` you can even improve on that and have a better Python interface with `aaargs`. If you want any of the nice features that are available through `click` or `typer` this is something that `aaargs` will never have. Personally, I mostly use `typer`, primarly because I have nested commands like `dask4dvc run --address 127.0.01:1234` or `dask4dvc repro --parallel`.

Rewriting the `mace` interface with `click` or `typer` would certainly be a bigger effort compared to this PR.
I also have to mention, that I've written the `aaargs` library so I'm totally biased here. But it has extensive testing, CI and everything so you don't have to worry about it breaking.

PR #46, opened 2022-11-11
partial bugfixes suggested by ilyes.

wcwitt:


wcwitt:
Closing because we're doing something else instead

PR #50, opened 2022-11-20
torchscript tweaks.

wcwitt:


ilyes319:
This adds info to the buffer (because torchscript can not read well buffers of submodules).

PR #56, opened 2022-11-29
update installation instructions in readme

stenczelt:
Updated the installation instructions in the readme for new users, so they use the correct conda channel for pytorch as well as CUDA 11.6. Added the installation of main dependencies with conda as well there.

see  https://pytorch.org/blog/pytorch-enterprise-support-update/ for information about deprecation


stenczelt:
@ilyes319 @davkovacs do you want the same for the main `brach` as well?

PR #70, opened 2023-02-06
Weights and Biases support

davkovacs:
This pull requests implements the option to track experiments using Weights and Biases as an optional dependency. 

Other small enhancements are specifying hidden irreps via `num_channels` and` max_L` and logging of average number of neighbours precisely.

### DIFF
--- a/scripts/run_train.py
+++ b/scripts/run_train.py
@@ -204,6 +204,17 @@ def main() -> None:
 
     # Build model
     logging.info("Building model")
+    if args.num_channels is not None and args.max_L is not None:
+        assert args.num_channels > 0, "num_channels must be positive integer"
+        assert args.max_L >= 0 and args.max_L < 4, "max_L must be between 0 and 3, if you want to use larger specify it via the hidden_irrpes keyword"
+        args.hidden_irreps = "{}x0e".format(args.num_channels)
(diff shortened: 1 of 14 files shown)


scripts/run_train.py
```diff
@@ -452,6 +465,19 @@ def main() -> None:
     logging.info(f"Number of parameters: {tools.count_parameters(model)}")
     logging.info(f"Optimizer: {optimizer}")
 
+    if args.wandb:
+        logging.info("Using Weights and Biases for logging")
+        wandb_config = {}
+        args_dict = vars(args)
```
RokasEl, line 471: maybe save this args dict as a json?

ilyes319: what is the use case of that?

davkovacs: I have added it to the wandn summary dict. Basically it is just easier to parse compared to the one-liner at the top of the log file, but contains the same information.

ilyes319: I see, we could also make this one liner a bit more readable. it get saves to the wandb forlder?

davkovacs: yes, it is in the `wandb-summary.json` file under the keyword `params`

Issue #68, opened 2023-01-23
Return learned local embedding


davkovacs:
It could be a useful addition to add a keyword to the model whcih would return the learned local embedding of the model. 

This could be just the invariant part of it, which is a tensor of shape `[N_atoms x (num_interactions x N_channels)]`. 

To implement we just need to return the invariant features after each interaction that are passed to the readout blocks. 

This would probably only give a good descriptor with a trained model because of normalisation.

PR #71, opened 2023-02-06
Added a way to extract invariant node embeddings

RokasEl:
[Related issue](https://github.com/ACEsuit/mace/issues/68)

Added a method `get_node_invariant_descriptors` to extract the invariant part of the learnt node representations using MACE and models that inherit the MACE class.

Cleaned up the forward by putting the calculation of initial embeddings in a separate function.

Added tests for the descriptor retrieval.

Cleaned up test_models.py by adding fixtures to reduce code repetition.

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -249,3 +249,36 @@ class MACECalculator(Calculator):
                     .cpu()
                     .numpy()
                 )
+    def get_invariant_descriptors(self, atoms=None) -> np.ndarray|list[np.ndarray]:
+        """ Extracts the invariant part from the node features after each interaction block in the MACE model.
+        :param atoms: ase.Atoms object
+        :return: np.ndarray (num_atoms, num_interactions, invariant_features) of invariant descriptors if num_models is 1 or list[np.ndarray] otherwise
+        """
+        if atoms is None and self.atoms is None:
+            raise ValueError("atoms not set")
+        elif atoms is None:
+            atoms = self.atoms
+            
+        if self.model_type != "MACE":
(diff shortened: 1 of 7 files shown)


ilyes319:
Looks great, for sure we need to add a readme part for that at least.

davkovacs:
We could have a function initialising a model with random weights and normalisation giving features on the same scale for all body-orders? And the user would just need to specify the model config and the chemical elements? 

This is a slightly different use-case from extracting the features from a trained model.

ilyes319:
The average number of neighbours would make a global rescaling. If you want to make all the body order on the same scale, you must rescale them individually. The average number of neighbours would be enough to guarantee none is too small.

gabor1:
> Looks great!
> 
> One thing to consider adding is an API for using it with a trained model. I am thinking about a thin wrapper class which takes a trained MACE model and returns an object that takes as input ase Atoms objects and returns the invariant features, perhaps in the arrays field of the Atoms object itself.
> 
> @gabor1 What do you think? How would the API be the most useful?

yes I think that would be very useful. it would be great to supply a "basic" descriptor trained on something big (at least element-diversity wise)

RokasEl:
Added `get_invariant_descriptors` method to the `MACECalculator` which allows to directly extract the features from an atoms object.

Also added more tests the calculator, feature extraction block, and the function itself attached onto the MACE class.

ilyes319:
@RokasEl I will be closing as implemented in the PR 0a74b7acd3680777c3ac1d8da4288160a386679d. Made a simpler and more general version.

PR #78, opened 2023-02-13
Committee error

LarsSchaaf:
# Committee error

This PR provides:
1. A mace calculator for using mace committees as a metric for model error
    - A list of mace models can be supplied to a given calculator. The calculator returns the variance between the predictions in energy in forces as a metric for the predicted model error. 
    - This method has been tested for committees, consisting of mace models trained on different random train/validation splits. 
2. A script for active learning using MD
    - As an example use case of the calculator a quick AL script is included. This simply runs MD until the predicted error reaches a set threshold.

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -11,6 +11,9 @@ from ase.stress import full_3x3_to_voigt_6_stress
 
 from mace import data
 from mace.tools import torch_geometric, torch_tools, utils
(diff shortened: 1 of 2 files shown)


ilyes319:
Looks good to me! @LarsSchaaf Can you provide a test following the calculator https://github.com/ACEsuit/mace/blob/main/tests/test_calculator.py ?

PR #83, opened 2023-02-17
no gradient with respect to model params in eval

davkovacs:


davkovacs:
I have also fixed a small bug in patience when there is no swa.

ilyes319:
There is a bug when predicting forces on isolated atoms. If the graph has no edges, or an atom has no edges, positions will never enter its construction so the energy does not have a backward computational graph. This needs a fix.

davkovacs:
Shall we merge this? I think all tests pass now.

PR #86, opened 2023-03-07
Argparse list arguments

gelzinyte:
Previous `type=list` would parse option

`--wandb_log_hypers "aa bb" ` -> `["a", "a", " ", "b", "b"]`

Now 

`--wandb_log_hypers aa bb` ->` ["aa", "bb"]`.

### DIFF
--- a/mace/tools/arg_parser.py
+++ b/mace/tools/arg_parser.py
@@ -460,7 +460,8 @@ def build_default_arg_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--wandb_log_hypers",
         help="The hyperparameters to log in Weights and Biases",
-        type=list,
+        type=str,
+        nargs="+",
         default=[
             "num_channels",
             "max_L",


davkovacs:
I think the expectation is that you specify the option as 
`--wandb_log_hypers="['aa', 'bb']"` not  `--wandb_log_hypers= 'aa', 'bb'`

So I am not sure this is a bug.

gelzinyte:
That still gets parsed as a string. 

For example, script: 
```
import argparse

parser = argparse.ArgumentParser()
parser.add_argument('--list-type', type=list)

for _, value in parser.parse_args()._get_kwargs():
    if value is not None:
        print(value)
```
gives
```
$ python parse.py --list-type "['aa', 'bb']"
['[', "'", 'a', 'a', "'", ',', ' ', "'", 'b', 'b', "'", ']']
```

ilyes319:
@davkovacs @gelzinyte Do you have settled on anything? I have no opinion on this. I think the point of parsing it as a string it to then infer the list using a python lib so the parsing in cleaner.

davkovacs:
I think the current convention is compatible with many other arguments being parsed in MACE. So we would need to either change all of them or none.

PR #88, opened 2023-03-08
in lammps docs, ensure model is cpu model.

wcwitt:
Hi @ilyes319, does this seem sensible to you? I assume it does nothing if the model is already cpu?


ilyes319:
Yes I think that should solve some problems, I merge.

PR #91, opened 2023-03-16
Update utils.py

sandipde:
at present the code crash if it contains a atoms with len 1 but without config_type key in info. Proposinf Fix  by introducing try: except.

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -5,15 +5,16 @@
 ###########################################################################################
 
 
+from glob import glob
+from typing import Union
+
+import numpy as np
 import torch
 from ase.calculators.calculator import Calculator, all_changes
 from ase.stress import full_3x3_to_voigt_6_stress
 
 from mace import data
 from mace.tools import torch_geometric, torch_tools, utils
-from typing import Union
-from glob import glob
-import numpy as np
 
 
 class MACECalculator(Calculator):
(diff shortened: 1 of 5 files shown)


ilyes319:
Looks good to me, we should probably add a warning when we do the expect. Something like "Isolated atom found with no config type, assuming isolated atom". Maybe there is a case for the expect to return False also and assume that the true isolated atom are only with the proposer config_type.

sandipde:
I was first confused but then realised It was a typo when I made the edit on web. Thanks for checking. I have updated the commit. Let me know if the proposal makes sense.

ilyes319:
That is why it defaults to false, check the updated code

gabor1:
In GAP, we pick up an “isolated atom” if it is (i) by itself in the structure, and (ii) the cell is larger than the receptive field, i.e. a sphere with the radius of the receptive field fits into the cell.


-- Gábor

Gábor Csányi
Professor of Molecular Modelling
Engineering Laboratory, University of Cambridge
Pembroke College Cambridge

Pembroke College supports CARA. A Lifeline to Academics at Risk.  http://www.cara.ngo/

On 29 Mar 2023, at 13:46, davkovacs ***@***.***> wrote:



What if I have a primitive cell? That should not be treated as isolated atom, I think that's why it was like that.

PR #92, opened 2023-03-24
Interaction layer specific settings

LarsSchaaf:
This commit allows for the cutoff and correlation order to be different for different message passing layers. 

`correlation` and `r_max` now accepts a list of length `num_interactions`. 

eg. 

```
--num_interactions=3
--correlation="[3,3,1]"
--r_max="[7.0,3.0,7.0]"
```

### DIFF
--- a/mace/data/atomic_data.py
+++ b/mace/data/atomic_data.py
@@ -7,6 +7,7 @@
 from typing import Optional, Sequence
 
 import torch.utils.data
+import numpy as np
 
 from mace.tools import (
     AtomicNumberTable,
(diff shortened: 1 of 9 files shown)


ilyes319:
@LarsSchaaf What do you think is needed for the clean up before merging this?

davkovacs:
Should we implement as part of tis PR the central atom dependent cutoff?

ilyes319:
yes that's a good place

davkovacs:
We were discussing the implementation of layer specific number of channels. 

Would this be a good place to put it? Is this PR planned to be merged in the near future?

ilyes319:
yes it is a good place for a fork to develop that

Issue #93, opened 2023-03-26
training with MPS - ARM64 Mac


stenczelt:
**Describe the bug**
Training of model cannot be initialised on M2 Mac, using MPS acceleration. Since apple GPUs don't support 64 bit floats, so one needs to set `default_dtype=float32` which is likely the issue.

**To Reproduce**
1. Try training a model with `--device=mps --default_dtype=float32` 

**Expected behavior**
The training should "just work" like elsewhere or on CPU.

**Desktop (please complete the following information):**
 - OS: MacOS, 13.2.1
 - M2 chip
 - Torch 2.0.0, 
 - Python 3.9

**Additional context**

training args used:

```
python ../scripts/run_train.py \
    --name="MACE_model" \
    --train_file="Al2O3_train.xyz" \
    --valid_fraction=0.05 \
    --test_file="Al2O3_test.xyz" \
    --config_type_weights='{"Default":1.0}' \
    --model="MACE" \
    --hidden_irreps='16x0e + 16x1o' \
    --r_max=5.0 \
    --batch_size=10 \
    --max_num_epochs=1500 \
    --swa \
    --start_swa=1200 \
    --ema \
    --ema_decay=0.99 \
    --amsgrad \
    --restart_latest \
    --device=mps \
    --default_dtype=float32
```

output:
```
2023-03-26 08:04:10.863 INFO: MACE version: 0.2.0
[...]

PR #94, opened 2023-03-26
remove explicit usage of float64

stenczelt:
For #93 

Incomplete, raises other issues


davkovacs:
Thank you, it has been resolved.

PR #95, opened 2023-03-31
ASE calculator refactor and jit dipole models

davkovacs:


davkovacs:
> We should add the variance and mean to the properties for writing issues.

Done in the above commit, I think it is ready to merge (if you verify that there is not performance difference to the previous calculator.

ilyes319:
Better to compute the mean and variance of the forces on the GPU and then transfer that one time to cpu.

bernstei:
Note that there are things like HAL that need the full list of forces and/or energies to compute the bias.

ilyes319:
That's a good point.  Maybe add a flag to enable it.

davkovacs:
After discussing with Cas we agreed that it is sufficient to return the mean and variance of energies and forces, it shouldn't be necessary to return the output force for all models in a committee for HAL.

davkovacs:
Thanks @bernstei I clarified again with Cas and have changed to return all committee forces.

PR #97, opened 2023-04-02
Refactor MACE subclasses - reduce code duplication & clearer logic

stenczelt:
Supersedes #65, implements the JIT part of #95 as well for free

Refactored subclasses of `MACE`:
- `ScaleShiftMACE`: Only scales & shifts MACE
- `AtomicDipolesMACE`: Only defined dipole calculation capabilities
- `EnergyDipolesMACE`: Energy & dipole calculation

Main changes:
- `__init__` is basically the same, there is only a little difference, which is handled with
    - readout block classes are kept in as class members
    - irreps of the last layer is computed with a function (overwritten by subclasses)
- parts of the forward pass calculation are separated into internal functions

### DIFF
--- a/mace/modules/models.py
+++ b/mace/modules/models.py
@@ -151,6 +151,25 @@ class MACE(torch.nn.Module):
             else:
                 self.readouts.append(LinearReadoutBlock(hidden_irreps))
 
+    def _forward_calculate_e0(self, data: Dict[str, torch.Tensor]) -> torch.Tensor:
+        """Forward pass internal - e0 calculation
+
+        Notes
+        -----
+        This is really a lookup from the e0 list we have for each node's element.
+
+        Parameters
+        ----------
+        data
+
+        Returns
+        -------
+        e0
+            shape: [n_nodes, ]
+
+        """
+        return self.atomic_energies_fn(data["node_attrs"])
+
     def forward(
         self,
         data: Dict[str, torch.Tensor],
@@ -184,7 +203,7 @@ class MACE(torch.nn.Module):
             )
 
         # Atomic energies
-        node_e0 = self.atomic_energies_fn(data["node_attrs"])
+        node_e0 = self._forward_calculate_e0(data)
         e0 = scatter_sum(
             src=node_e0, index=data["batch"], dim=-1, dim_size=num_graphs
         )  # [n_graphs,]
@@ -276,8 +295,8 @@ class ScaleShiftMACE(MACE):
         compute_displacement: bool = False,
     ) -> Dict[str, Optional[torch.Tensor]]:
         # Setup
-        data["positions"].requires_grad_(True)
         data["node_attrs"].requires_grad_(True)
+        data["positions"].requires_grad_(True)
         num_graphs = data["ptr"].numel() - 1
         displacement = torch.zeros(
             (num_graphs, 3, 3),
--- a/mace/modules/models.py
+++ b/mace/modules/models.py
@@ -36,6 +36,8 @@ from .utils import (
 
 @compile_mode("script")
 class MACE(torch.nn.Module):
+    """MACE model"""
+
     def __init__(
         self,
         r_max: float,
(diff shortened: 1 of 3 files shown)


stenczelt:
@davkovacs there's a more general refactor in https://github.com/stenczelt/mace/tree/ENH/refactor-model-v3 where I have separated out the backbone of the MACE model (agnostic of what you're calculating) and the quantity-specific readout blocks and gradient calculations. 
That one is a somewhat larger refactor, though I think it's quite neat. Shall I update this PR to that instead?

davkovacs:
I would say let’s have one refactor and test that thoroughly

stenczelt:
Roger that @davkovacs !

mace/modules/core_models.py
```diff
@@ -0,0 +1,378 @@
...
+    def __init__(
+        self,
+        r_max: float,
+        num_bessel: int,
+        num_polynomial_cutoff: int,
+        max_ell: int,
+        interaction_cls: Type[InteractionBlock],
+        interaction_cls_first: Type[InteractionBlock],
+        num_interactions: int,
+        num_elements: int,
```
stenczelt, line 58: The number of elements is redundant here. I propose removing it. ok?

mace/modules/core_models.py
```diff
@@ -0,0 +1,378 @@
...
+            class to be used for the first layer's interaction block
+        num_interactions
+            number of interaction layers to use
+        num_elements
+            redundant parameter for the number of elements
+        hidden_irreps
+            hidden irreducible representations, basically the size of the layer features
+            and hence direct control on the size of the model
+        MLP_irreps
+        avg_num_neighbors
```
stenczelt, line 91: These docstrings are duplicated across the classes exposed to the user as well. I think this is fine, so whichever one you are using you see the docstring directly. If you disagree feel free to remove the redundant ones and perhaps keep them for the main one

mace/modules/core_models.py
```diff
@@ -0,0 +1,378 @@
...
+        interaction_cls_first: Type[InteractionBlock],
+        num_interactions: int,
+        num_elements: int,
+        hidden_irreps: o3.Irreps,
+        MLP_irreps: o3.Irreps,
+        avg_num_neighbors: float,
+        atomic_numbers: List[int],
+        correlation: int,
+        gate: Optional[Callable],
+        radial_MLP: Optional[List[int]] = None,
```
stenczelt, line 65: I have not changed this for the sake of consistency, method parameters are conventionally lower case, so I propose changing these `MLP` parameters to lowercase

mace/modules/models.py
```diff
@@ -254,120 +187,364 @@ def forward(
...
+        **kwargs
+        """
         super().__init__(**kwargs)
-        self.scale_shift = ScaleShiftBlock(
-            scale=atomic_inter_scale, shift=atomic_inter_shift
-        )
+
+
+@compile_mode("script")
+class AtomicDipolesMACE(MaceCoreModel, DipoleModelMixin):
```
stenczelt, line 245: why is this called "AtomicDipolesMACE"? Should it not just be `DipolesMACE`

mace/modules/models.py
```diff
@@ -254,120 +187,364 @@ def forward(
...
+    def _last_layer_irreps(hidden_irreps) -> o3.Irreps:
+        """dipole: Select only l=1 vectors for last layer"""
+        return o3.Irreps(str(hidden_irreps[1]))
 
     def forward(
         self,
         data: Dict[str, torch.Tensor],
         training: bool = False,
-        compute_force: bool = True,
+        compute_force: bool = False,
```
stenczelt, line 307: If this class cannot calculate forces, why don't we remove the associated parameters?  Are they intended for compatibility with wrappers?

mace/modules/models.py
```diff
@@ -254,120 +187,364 @@ def forward(
...
 
-        # Sum over nodes in graph
-        inter_e = scatter_sum(
-            src=node_inter_es, index=data["batch"], dim=-1, dim_size=num_graphs
-        )  # [n_graphs,]
+    def forward(
+        self,
+        data: Dict[str, torch.Tensor],
+        training: bool = False,
+        compute_force: bool = True,
```
stenczelt, line 420: Same here, parameters as traps only. Can understand if it is compatibility and perhaps will be implemented later on.

mace/modules/models.py
```diff
@@ -254,120 +187,364 @@ def forward(
...
-            "stress": stress,
-            "displacement": displacement,
+            "dipole": self._calc_total_dipole(data, atomic_dipoles),
+            "atomic_dipoles": atomic_dipoles,
         }
 
-        return output
+
+@compile_mode("script")
+class ScaleShiftEnergyDipoleMACE(EnergyDipolesMACE, ScaleShiftEnergyModelMixin):
```
stenczelt, line 489: this is just an extra, which was easy to add due to the separated logic of dipoles & energy

tests/test_models.py
```diff
@@ -194,3 +205,38 @@ def test_energy_dipole_mace():
...
+    output = model(
+        data_batch_1,
+        training=True,
+    )
+
+    for key in ["energy", "node_energy", "forces", "dipole", "atomic_dipoles"]:
+        assert key in output
+
+
+def test_scaled_and_shifted(dipole_model_config, data_batch_1):
```
stenczelt, line 225: While the tests were enormously useful while refactoring, there is more to test on all these classes. Should be somewhat easier now.

stenczelt:
@davkovacs @ilyes319 I have added my comments relating to what I have done, you might be able to answer these or make the appropriate judgement calls. Please let me know if you need clarification on anything or want changes which I should make

PR #100, opened 2023-04-10
Fix unweighted EnergyForcesLoss

chaitjo:
This PR fixes unweighted EnergyForcesLoss.

This is a trivial fix - it seems that the code for WeightedEnergyForcesLoss was copied for the unweighted default variant, EnergyForcesLoss.

Currently, EnergyForcesLoss is being used as a default fallback here:
https://github.com/ACEsuit/mace/blob/44b6da4a5edaa4c3ef867a11728555403d1d475d/scripts/run_train.py#L180

### DIFF
--- a/mace/modules/loss.py
+++ b/mace/modules/loss.py
@@ -78,21 +78,11 @@ def weighted_mean_squared_error_dipole(ref: Batch, pred: TensorDict) -> torch.Te
 
 
 class EnergyForcesLoss(torch.nn.Module):
-    def __init__(self, energy_weight=1.0, forces_weight=1.0) -> None:
+    def __init__(self, ) -> None:
         super().__init__()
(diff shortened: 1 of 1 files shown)


davkovacs:
Thanks for spotting this! I would say we don't need a separate class for this case. How about just initialising the weighted one with weights being 1.0 ?

chaitjo:
I updated the loss modules.

An alternative formulation from the current one in the PR could be to get rid of the EnergyForcesLoss class altogether, and initialise the default setting in run_train using the WeighterEnergyForcesLoss with 1.0/1.0 weights.

ilyes319:
I think that's the best, to remove it altogether

davkovacs:
I think it is ready to merge

PR #117, opened 2023-06-15
Slight modification to HDF5Dataset

sivonxay:
If a `h5py.File` object is an attribute of the Dataset, pickle will complain that the file is not pickleable. 

This change resolves that by making file a property that is constructed and cached when it is first accessed. This allows the `h5py.File` to be excluded when pickling.

### DIFF
--- a/mace/data/hdf5_dataset.py
+++ b/mace/data/hdf5_dataset.py
@@ -30,7 +30,6 @@ class HDF5ChainDataset(ChainDataset):
         return _d
 
     def __call__(self):
-        self.file = h5py.File(self.file, "r")
         datasets = []
         for i in range(self.length):
             grp = self.file["config_" + str(i)]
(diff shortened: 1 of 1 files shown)


ilyes319:
Looks good to me. Do you have a good example of this behaviour to test it?

sivonxay:
Yeah, here's an example that will trigger the error. I can't attach the h5 file here, but it should be the same for any hdf5 file. Let me know if you want this specific one.

```
from mace.tools.torch_geometric import DataLoader
from mace.data import HDF5Dataset
from mace import tools

z_table = tools.AtomicNumberTable(list(range(18))[1:])
dataset = HDF5Dataset('chem_h5/processed_chem/train_0.h5', z_table=z_table, r_max=5.0)
loader = DataLoader(dataset, batch_size=32, shuffle=True, num_workers=0)
# Get one batch from the dataloader. This works fine
batch = next(iter(loader))

# Now we make a dataloader with 4 workers
loader = DataLoader(dataset, batch_size=32, shuffle=True, num_workers=4)
# With 4 workers, it now throws an error
next(iter(loader))
```

The error thrown is: `TypeError: h5py objects cannot be pickled`

ilyes319:
Great thank you I am merging.

PR #133, opened 2023-07-17
Multi-processing

mavaylon1:


scripts/preprocess_data.py
```diff
@@ -6,10 +6,17 @@
 import numpy as np
 import json
 import random
+import tqdm
+from glob import glob
+
 import h5py
 
+from joblib import Parallel, delayed
```
sivonxay, line 14: Parallel and delayed aren't used. These imports should be removed

scripts/preprocess_data.py
```diff
@@ -108,70 +179,101 @@ def main():
...
+        i.join()
+
+    logging.info("Computing statistics")
+    if len(atomic_energies_dict) == 0:
+        atomic_energies_dict = get_atomic_energies(args.E0s, collections.train, z_table)
+    atomic_energies: np.ndarray = np.array(
+        [atomic_energies_dict[z] for z in z_table.zs]
+    )
+    logging.info(f"Atomic energies: {atomic_energies.tolist()}")
+    _inputs = ['/pscratch/sd/m/mavaylon/processed_chem', z_table, args.r_max, atomic_energies, args.batch_size]
```
sivonxay, line 210: This hard coded path should be removed

scripts/preprocess_data.py
```diff
@@ -20,6 +27,69 @@
...
+    
+    avg_num_neighbors, mean, std = compute_statistics(train_loader, atomic_energies)
+
+    output = [avg_num_neighbors, mean, std]
+    return output
+
+
+def pool_compute_stats(inputs): #inputs = (path_to_files, z_table, r_max, atomic_energies, batch_size)
+    path_to_files, z_table, r_max, atomic_energies, batch_size = inputs
+    pool = mp.Pool(processes=int(os.cpu_count()/4))
```
sivonxay, line 84: This should use the `num_process` obtained from the parser

scripts/preprocess_data.py
```diff
@@ -108,70 +179,101 @@ def main():
...
         random.shuffle(collections.train)
 
-    with h5py.File(args.h5_prefix + "train.h5", "w") as f:
-        # split collections.train into batches and save them to hdf5
-        split_train, drop_last = split_array(collections.train, args.batch_size)
-        f.attrs["drop_last"] = drop_last
-        for i, batch in enumerate(split_train):
-            save_configurations_as_HDF5(batch, i, f)
+    # split collections.train into batches and save them to hdf5
+    split_train = np.array_split(collections.train,int(os.cpu_count()/4))
```
sivonxay, line 183: This should also use the `num_process` from the parser

scripts/preprocess_data.py
```diff
@@ -108,70 +179,101 @@ def main():
...
+    drop_last = False
+    if len(collections.train) % 2 == 1:
```
sivonxay, line 185: Why do we only drop last when the length of the initial collection is an odd number? I would think that we need to check if `(len(collections.train) // num_process) % batch_size != 0` or `((len(collections.train) // num_process) +1 ) % batch_size != 0` (The latter in the case in which `collections.train` is not evenly divisible into `num_process` sub-arrays)

scripts/preprocess_data.py
```diff
@@ -108,70 +179,101 @@ def main():
...
-        }
-        del train_dataset
-        del train_loader
-        with open(args.h5_prefix + "statistics.json", "w") as f:
-            json.dump(statistics, f)
     
     logging.info("Preparing validation set")
     if args.shuffle:
         random.shuffle(collections.valid)
+    split_valid = np.array_split(collections.valid, int(os.cpu_count()/4))
```
sivonxay, line 235: `num_process`

scripts/preprocess_data.py
```diff
@@ -108,70 +179,101 @@ def main():
...
-        f.attrs["drop_last"] = drop_last
-        for i, batch in enumerate(split_valid):
-            save_configurations_as_HDF5(batch, i, f)
+    def multi_valid_hdf5(process):
+        with h5py.File(args.h5_prefix + "valid_" + str(process)+".h5", "w") as f:
+            f.attrs["drop_last"] = drop_last
+            save_configurations_as_HDF5(split_valid[process], process, f)
+    
+    processes = []
+    for i in range(int(os.cpu_count()/4)):
```
sivonxay, line 246: `num_process`

scripts/preprocess_data.py
```diff
@@ -108,70 +179,101 @@ def main():
...
         for name, subset in collections.tests:
-            with h5py.File(args.h5_prefix + name + "_test.h5", "w") as f:
-                split_test, drop_last = split_array(subset, args.batch_size)
-                f.attrs["drop_last"] = drop_last
-                for i, batch in enumerate(split_test):
-                    save_configurations_as_HDF5(batch, i, f)
+            drop_last = False
+            if len(subset) % 2 == 1:
+                drop_last = True
+            split_test = np.array_split(subset, int(os.cpu_count()/4))
```
sivonxay, line 265: `num_process`

scripts/preprocess_data.py
```diff
@@ -108,70 +179,101 @@ def main():
...
-                f.attrs["drop_last"] = drop_last
-                for i, batch in enumerate(split_test):
-                    save_configurations_as_HDF5(batch, i, f)
+            drop_last = False
+            if len(subset) % 2 == 1:
+                drop_last = True
+            split_test = np.array_split(subset, int(os.cpu_count()/4)) 
+
+            processes = []
+            for i in range(int(os.cpu_count()/4)):
```
sivonxay, line 268: `num_process`

scripts/preprocess_data.py
```diff
@@ -108,70 +179,101 @@ def main():
...
+                p = mp.Process(target=multi_test_hdf5, args=[i, name])
+                p.start()
+                processes.append(p)
 
+            for i in processes:
+                i.join()
+    finish = time.perf_counter()
+    print(f'Finished in {round(finish-start, 2)} second(s)')
```
sivonxay, line 276: This should go into the logger or removed

mace/tools/arg_parser.py
```diff
@@ -543,6 +544,12 @@ def build_preprocess_arg_parser() -> argparse.ArgumentParser:
         default=None,
         required=False,
     )
+    parser.add_argument(
+        "--num_process",
+        help="The user defined number of processes to use",
```
sivonxay, line 549: Also add to this comment that this implicitly sets the # of files that will be created for the train, test, and val. Possibly a discussion/warning that if the dataset size is too large/ very small this may need to be tweaked independent of the number of CPUs?

Actually, we may want to consider decoupling the two and having a `num_process` and `num_files` argument?

sivonxay (commented):
I think this looks good. Just a few minor notes and suggestions.

scripts/preprocess_data.py
mavaylon1: I added the argument to the parser and it cannot find it when running 
```
python ./mace/scripts/preprocess_data.py \
    --train_file="/pscratch/sd/m/mavaylon/mace_demo/t1x_compund_subset.xyz" \
    --test_file="/pscratch/sd/m/mavaylon/mace_demo/t1x_compund_subset.xyz" \
    --num_process = 64 \
    --valid_fraction=0.05 \
    --atomic_numbers="[1, 6, 7, 8]" \
    --r_max=4.5 \
    --h5_prefix="/pscratch/sd/m/mavaylon/processed_chem/" \
    --E0s="average" \
    --seed=123 \
```

scripts/preprocess_data.py
mavaylon1: The way we have it setup the batch_size doesn't matter. We are just separating the data based on cpu count and checking the drop_last is based on the odd number in split_array.

mavaylon1:
@ilyes319 I added the number of processes as an argument to the parser, but it does not recognize it when running 

```
python ./mace/scripts/preprocess_data.py \
    --train_file="/pscratch/sd/m/mavaylon/mace_demo/t1x_compund_subset.xyz" \
    --test_file="/pscratch/sd/m/mavaylon/mace_demo/t1x_compund_subset.xyz" \
    --num_process=64 \
    --valid_fraction=0.05 \
    --atomic_numbers="[1, 6, 7, 8]" \
    --r_max=4.5 \
    --h5_prefix="/pscratch/sd/m/mavaylon/processed_chem/" \
    --E0s="average" \
    --seed=123 \
```

I can't seem to figure out what I am. missing here.

mavaylon1:
> Author

Figured it out.

mace/tools/arg_parser.py
mavaylon1: I added the note in the help.

scripts/preprocess_data.py
sivonxay: Correct me if i'm wrong, but I think `drop_last` is not primarily used in the chunking of the data, rather it's an argument to the dataloader which tells the pytorch dataloader whether to always return uniform batch sizes (and therefore drop the last remainders)?

mavaylon1: I am not sure, but the logic to set drop last was based off of the input being divisible by 2 in the original. I took that and forewent the rest of split_array.

mavaylon1: def split_array(a: np.ndarray, max_size: int):
    drop_last = False
    if len(a) % 2 == 1:
        a = np.append(a, a[-1])
        drop_last = True

mavaylon1:
@sivonxay could you take another look?
Also having
```
from joblib import Parallel, delayed
```
sems to be required but I am not sure how. Without this import, it slows down heavily.

sivonxay:
I don't see why importing joblib would change the performance. Can you try the test in a clean environment without joblib installed?

mace/data/hdf5_dataset.py
```diff
@@ -153,6 +155,13 @@ def __getitem__(self, index):
         )
         return atomic_data
 
+def dataset_from_sharded_hdf5(files, z_table, r_max):
```
sivonxay, line 158: Add type hints

scripts/preprocess_data.py
```diff
@@ -21,6 +27,36 @@
 from mace.modules import compute_statistics
 
 
+compute_stats_results = []
+
+def compute_stats_callback(result):
+    compute_stats_results.append(result)
+
+def compute_stats_target(file, z_table, r_max, atomic_energies, batch_size):
```
sivonxay, line 35: Add type hints

mavaylon1:
@ilyes319 Hey Ilyes github is not letting me add you to review. Could you take a look for a merge?

PR #143, opened 2023-08-14
Implementation of Huggingface/Accelerate

pbenner:
Dear MACE authors,

the current implementation of the multi-GPU branch requires slurm, which is not available on every GPU cluster (for instance ours). Huggingface/Accelerate is an alternative, which is easy to use and available as a simple python package. This pull request provides a basic implementation. The logging and your custom checkpointing must still be adapted, but it should be quite easy.

Once implemented, multi-GPU training can be executed using:
```bash
BASEDIR=data-train-preprocessed

accelerate launch --multi_gpu ./mace/scripts/run_train.py \
    --name="MACE_model" \
    --train_file="./${BASEDIR}/train.h5" \
    --valid_file="./${BASEDIR}/valid.h5" \
    --statistics_file="./${BASEDIR}/statistics.json" \
    --accelerate \
    --num_workers=16 \
    --model="ScaleShiftMACE" \
    --num_interactions=2 \
    --num_channels=128 \
    --max_L=1 \
    --correlation=3 \
    --batch_size=32 \
    --valid_batch_size=32 \
    --max_num_epochs=100 \
    --swa \
    --start_swa=60 \
    --ema \
    --ema_decay=0.99 \
    --amsgrad \
    --error_table='PerAtomMAE' \
    --device=cuda \
    --seed=42
```
Notice the new *--accelerate* flag as an alternative to the *--distributed* option.

Hope this is of interest! Best, Pb


ilyes319:
Hey Philipp,

Thank you very much for your PR, this looks great. What kind of changes are missing from your PR?

pbenner:
Hi Ilyes,

Accelerate is quite simple, it launches several processes and handles distributes training by encapsulating the model, dataloader, trainer, and scheduler objects in a custom accelerator object. These parts are open:

- Make sure that only the main process is logging. There is a logger provided by accelerate that checks if the main process calls the logger
- Only the main process should save the model. I've implemented that, but didn't go through the entire SWA / Checkpoint code. You would also have to decide if you want to save the encapsulated objects or just the torch objects. Accelerate provides some simple code for saving all encapsulated objects (model, trainer etc.)
- If you require the accelerate package also for single GPU usage, you can simplify the code (i.e. drop all the if statements for checking whether the accelerate object is None)

pbenner:
I implemented the remaining points, seems to be work nicely here. Best, Pb

ilyes319:
Amazing, thank you. I will review it asap. I will ask several people to try it so we are sure everything is correct with the checkpointing.

PR #155, opened 2023-09-14
Merging multi-GPU and h5 dataloader into develop.

samwaltonnorwood:
Integrates the changes for distributed training & h5 data processing from the current state of the multi-GPU branch. Note the changes for distributed evaluation are a bit invasive - we use torchmetrics to aggregate all error quantities so the associated functions (compute_mae, etc) have been refactored into the MACELoss(Metric) class - but the behavior on 1 GPU should be identical.

### DIFF
--- a/mace/data/__init__.py
+++ b/mace/data/__init__.py
@@ -8,8 +8,12 @@ from .utils import (
     config_from_atoms_list,
     load_from_xyz,
     random_train_valid_split,
+    save_configurations_as_HDF5,
     test_config_types,
+    save_dataset_as_HDF5,
+    save_AtomicData_to_HDF5,
 )
+from .hdf5_dataset import HDF5Dataset
 
 __all__ = [
     "get_neighborhood",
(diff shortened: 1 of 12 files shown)


samwaltonnorwood:
@ilyes319 here's the edit to run_train.py I mentioned - it applies for this PR and for the existing multi-GPU branch.

```python
    if args.compute_avg_num_neighbors:
        avg_num_neighbors = modules.compute_avg_num_neighbors(train_loader)
        if args.distributed:
            num_graphs = torch.tensor(len(train_loader.dataset)).to(device)
            num_neighbors = num_graphs * torch.tensor(avg_num_neighbors).to(device)
            torch.distributed.all_reduce(num_graphs, op=torch.distributed.ReduceOp.SUM)
            torch.distributed.all_reduce(num_neighbors, op=torch.distributed.ReduceOp.SUM)
            args.avg_num_neighbors = (num_neighbors / num_graphs).item()
        else:
            args.avg_num_neighbors = avg_num_neighbors
    logging.info(f"Average number of neighbors: {args.avg_num_neighbors}")
```

PR #178, opened 2023-09-28
Scripts entrypoints

### LINKED ISSUE #177: entry points for scripts
Would people be interested in adding entry points for some/all of the scripts, so they can become executables in the normal path when mace is pip installed?  If so, I'd be happy to have a go at it.  I'm hoping it's as simple as a few lines added to `setup.cfg`. There's a contribution of a mace fit wrapper to `wfl`, and it's easiest if we can count on `run_train.py` to in the path.

bernstei:
Move all python scripts in `scripts/` to `mace/cli/`, refactor what's needed so they all have `main` functions, and add them as `console_entrypoints` to `setup.cfg`

closes #177


ilyes319:
I would keep one copy of run train in scripts for backward compatibility. Maybe we can delete it in the future.

bernstei:
I think having two copies of supposedly the same script is just asking for drift.  Can we make the one in scripts a symlink? I'm pretty sure that git can handle that, and the pip install doesn't any scripts anyway, I think.

bernstei:
> Can we make the one in scripts a symlink?

Done now, and git seems fine with it.

bernstei:
@ilyes319 what's the status of this PR? I want to run a mace fit in the `wfl` CI, and for that I need to settle on where to look for `run_train.py`, so I'd like to know if I should do it based on the original file structure or the one I propose in this PR (i.e. if this PR will be merged soon).

ilyes319:
The missing parts are changing the readme and the doc to match the new location.

bernstei:
> The missing parts are changing the readme and the doc to match the new location.

I'll have a go at that later today.

bernstei:
I just pushed a change to the readme and a coupe of files in `tests/`, but didn't see any other docs to change.

ilyes319:
I have added direct wrapper. I think it is fine to keep the doc, but we really need to update when we merge to main.

bernstei:
> I have added direct wrapper. I think it is fine to keep the doc, but we really need to update when we merge to main.

I'd be happy to update the doc, I just didn't see anything that referred to those scripts except the README, which I did update, I thought.

ilyes319:
it is in User guide in the Training and Evaluation sections. With the wrapper, they are still valid so it is fine, but would be nice to make them better. Maybe we could add even another section about the client interface?

PR #239, opened 2023-11-28
Fix `mace_mp` download

janosh:
Was trying to download from 

```py
urllib.request.urlretrieve(model, cached_model_path)
```

which should have been

```py
urllib.request.urlretrieve(checkpoint_url, cached_model_path)
```

since `model` can be `None`. Thanks @utf for reporting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ mace_eval_configs \
 
 ## Tutorial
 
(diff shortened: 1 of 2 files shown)


ilyes319:
good catch

janosh:
Quick merge! :) Was going to add a test.

PR #268, opened 2023-12-23
Add an option to set the D3 damping function, which defaults to D3(BJ)

Andrew-S-Rosen:
This PR adds the `damping` parameter to the foundation model calculator, which defaults to "bj" for backwards compatibility (and because it makes the most sense). This is also done for added clarity, as it's not mentioned anywhere in the docstring whether a damping function is used. In fact, at a quick glance, the user might be misled to thinking no damping is used since "D3" is printed to screen even though it is actually D3(BJ).

### DIFF
--- a/mace/calculators/foundations_models.py
+++ b/mace/calculators/foundations_models.py
@@ -1,7 +1,7 @@
 import os
 import urllib.request
 from pathlib import Path
-from typing import Union
+from typing import Literal, Union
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
This might cause conflicts with #264. @ilyes319, feel free to add this to your PR if you wish.

PR #273, opened 2024-01-03
load_foundations support for max_L other than 2

bernstei:
Pass `max_L` argument to `load_foundations`, and use it to control j (# of contractions) loop


ilyes319:
looks great, I merge

PR #275, opened 2024-01-04
Add MACE-OFF23 models

davkovacs:
This PR implements the interface to the MACE-OFF23 models.


ilyes319:
@davkovacs I think we should also add a paragraph in the readme about the models

mace/calculators/foundations_models.py
```diff
@@ -121,6 +121,72 @@ def mace_mp(
     return calc
 
 
+def mace_off(
+    model: Union[str, Path] = None,
+    device: str = "",
+    default_dtype: str = "float32",
```
jthorton, line 127: The default dtype does not match the doc string.

davkovacs: good catch, I am updating it.

davkovacs:
Ready to be merged.

jthorton:
Might be good to add a test as well to make sure nothing gets broken in the future?

ilyes319:
You can use the same test as the mace_mp

davkovacs:
Can you test the tests please! @jthorton

ilyes319:
ready to merge?

mace/calculators/foundations_models.py
```diff
@@ -121,6 +121,79 @@ def mace_mp(
...
+            # download and save to disk
+            print(f"Downloading MACE model from {checkpoint_url!r}")
+            print(
+                f"By downloading the model you accept the ASL license, see https://github.com/gabor1/ASL"
+            )
+            urllib.request.urlretrieve(checkpoint_url, cached_model_path)
+            print(f"Cached MACE model to {cached_model_path}")
+        model = cached_model_path
+        msg = f"Using MACE-OFF23 MODEL for MACECalculator with {model}"
+        print(msg)
```
jthorton, line 174: Looks like we missed this print which should only trigger if we don't just want to return the model.

PR #280, opened 2024-01-07
Update foundation models

chiang-yuan:
- Update `mace-mp` model series `small`, `medium` and `large` to be consistent in loss weights with https://arxiv.org/abs/2401.00096. 
- Rename `small` model as `energy` for energy the fine-tuned one. 

@ilyes319 we may want to update this as soon as possible to avoid confusion for our users

### DIFF
--- a/mace/calculators/foundations_models.py
+++ b/mace/calculators/foundations_models.py
@@ -60,9 +60,10 @@ def mace_mp(
     elif model in (None, "small", "medium", "large") or str(model).startswith("https:"):
         try:
             urls = dict(
-                small="https://tinyurl.com/2jmmb8b7",  # 2023-12-10-mace-128-L0_energy_epoch-249.model (31M, 3.8M params)
+                small="http://tinyurl.com/2s3f6y9p",  # 2023-12-10-mace-128-L0_epoch-199.model (31M, 3.8M params)
                 medium="http://tinyurl.com/24z8czmf",  # 2023-12-12-mace-128-L1_epoch-199.model (42M, 4.7M params)
                 large="http://tinyurl.com/3a637wz2",  # 2024-01-07-mace-128-L2_epoch-199.model (61M, 5.7M params)
+                energy="https://tinyurl.com/2jmmb8b7",  # 2023-12-10-mace-128-L0_energy_epoch-249.model (31M, 3.8M params)
             )
             checkpoint_url = (
                 urls.get(model, urls["medium"])
--- a/mace/calculators/foundations_models.py
+++ b/mace/calculators/foundations_models.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from typing import Union
 
 import torch
+
 from ase import units
 from ase.calculators.mixing import SumCalculator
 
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
It would be beneficial to start uploading to pip and incrementing versions to prevent unexpected changes to end-users.

chiang-yuan:
@Andrew-S-Rosen you mean another package? `mace` is already used by other repository on PyPl

Andrew-S-Rosen:
It doesn't really matter if there is an existing package on PyPI. This happens all the time. It could be uploaded to PyPI as `mace-ml` or something similar, perhaps.

gabor1:
We can use mace-ff for now. Can be changed later

chiang-yuan:
@Andrew-S-Rosen understood. However there is an immediate necessity that we need to change the naming of models to confirm what we write in the preprint. During preprint preparation we discussed we all need to re-run the benchmark once we finalize the model, which we still haven't finalized the model yet so I am not sure the downstream application should be too sensitive about the change?

Andrew-S-Rosen:
FWIW, I'm not suggesting to hault the merge.

gabor1:
I thought everyone ran with the current medium model so no need to rerun?

ilyes319:
I made sure the medium model which is the one supposed to be used was up to date, so I do not understand why you want to update it?

ilyes319:
I am not sure having two small model is interesting. As long as we clearly say in the doc that the small one is the energy variant, I think it is fine.

chiang-yuan:
The loss weights are different. The medium model you select originally uses 1:1:10 for energy:forces:stress, and the small uses energy fine-tuned one 10:1:100. I don't hold strong opinion we need to rerun them all as long as they reproduce reasonable result and has the same hyperparams.

ilyes319:
I don't think them having the same loss weights is important as long as it is clearly said in the doc. They have different purpose and the idea was to provide a fast model that also had very good energetics.

gabor1:
by "fine tune" you mean the swa setting? clearly the current default medium model could also benefit from that, but at this point I don't think we want to rerun anything.

chiang-yuan:
I manually extended training with new loss weights, not swa originally
implemented.

Yes my point is we were saying we will rerun the models after finalizing
the model. If we decide to settle on suboptimal model that is fine too, but
we will need to revise training sections then (or not explicitly mentioning
loss weights in the text)

gabor1:
But if you only did this to the small model, that's probably not what we want to rerun everything with. or do you have an high-energy-weight continuation of all model sizes?

ilyes319:
I think there are 3 cases here to seperate: 
- the large model needs to be updated to be consistent
- the small model to me can stay the same as before, we just need to be more explicit that this is with the high energy weights
- I still do not get why the medium model needs to be changed. I was pretty sure I used the latest 199 epochs medium model as the latest checkpoint, so I do not get why it needs to be changed (maybe the date was confusing).

gabor1:
was the medium model used by everyone up to now trained either with SWA high energy weights or by a manual continuation with high energy weights? @chiang-yuan is implying that it was not.

chiang-yuan:
That's my confusion too. What are the loss weights we actually want? I didn't run other model sizes just because we briefly discussed about having small energy fine-tuned one on Slack. In principle 50 epochs for medium and large are all doable if we just want to provide it for end users (not rerunning benchmark examples if that is too much work?)

ilyes319:
no it was not, we tested that only for the small model

ilyes319:
@chiang-yuan I think the checkpoint is the same, I just forgot to update the date in the comment string.

chiang-yuan:
@gabor1 the one everyone uses now adopts the loss weights (1:1:10) - `2023-12-03-mace-128-L1_epoch-199.model`

PR #288, opened 2024-01-09
print validation error before any training

bernstei:
Refactor logging of validation error in tools/train.py so that it can be called before any training has taken place


bernstei:
I did this off foundations, but I don't know if it can easily be cherry-picked into main.

PR #290, opened 2024-01-11
Add link to github models

janosh:
Based on this answer https://stackoverflow.com/a/35160609 and my own testing, appending `?confirm=yTib` to the download URL avoids the Google Drive virus scan page that the model checkpoint links are automatically redirected to.

I confirmed locally that this resolves the pickle error reported by @Andrew-S-Rosen and also seen in [`matcalc`](https://github.com/materialsvirtuallab/matcalc/actions/runs/7486930755/job/20378280067):

```py
>       magic_number = pickle_module.load(f, **pickle_load_args)
E       _pickle.UnpicklingError: invalid load key, '<'.

/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/torch/serialization.py:1246: UnpicklingError
```

### DIFF
--- a/mace/calculators/foundations_models.py
+++ b/mace/calculators/foundations_models.py
@@ -57,8 +57,8 @@ def mace_mp(
     elif model in (None, "small", "medium", "large") or str(model).startswith("https:"):
         try:
             urls = dict(
-                small="https://tinyurl.com/2jmmb8b7",  # 2023-12-10-mace-128-L0_energy_epoch-249.model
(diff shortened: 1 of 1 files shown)


ilyes319:
My downloads are still failling locally with the same error.

janosh:
maybe print the URL to make sure it contains the query string

ilyes319:
When I click on the new url, it still opens the anti-virus page. It is weird because I bypassed that by providing a tinyurl to directly the url after the anti-virus. But somehow Google drive updated this url to include a check.

PR #302, opened 2024-01-19
Adding pre-commit checks and github action runners

hatemhelal:
This PR adds:
* pre-commit configuration to format code with `black`, `isort` and check lint with `pylint`
* github actions to run pre-commit and pytest unittests
* Additional `dev` install target to setup a virtualenv with required packages for checks


ilyes319:
Very nice, thank you @hatemhelal

PR #320, opened 2024-02-07
Add finetuning for maceoff models

jharrymoore:
Allow finetuning of maceoff models

Notes

- Slightly different way of initialising the models: using the parameter copying scheme from the mace-mp does not seem to work - since we want to keep the number of elements unchanged we simply intiialize from a checkpoint


ilyes319:
Supported in the main branch, thank you @jharrymoore

PR #352, opened 2024-03-26
Update plot_train.py

zhubonan:
Fix PDF file names of the `mace_plot_train` command. . Previous, the names of the output files are tuples:

```
('<my_filename>',).pdf
```

is this due to passing a list of grouped keys to the `groupby`, which results in a tuple of grouped keys being returned. 
Since we are only grouping a single key (`'name'`), passing a string should be sufficient.


ilyes319:
Thanks @zhubonan !

PR #365, opened 2024-04-02
Support torch.compile with ASE

hatemhelal:


ilyes319:
That is already supported in the main branch, thanks @hatemhelal

PR #370, opened 2024-04-05
fix rms/rmse for forces, off by a factor of 1/sqrt(3)

misko:
In the current implementation RMS/RMSE of forces is scaled by a factor of 1/sqrt(3) to where I think it should be, although I might be wrong. 

`return np.sqrt(np.mean(np.square(delta))).item()`

^ delta has shape (n,3) and n.mean() divides the sum out by n*3, when it should only be dividing by n, averaging over number of atoms, not 3*number of atoms. 

I checked a 
https://github.com/mir-group/pytorch_runstats/blob/main/torch_runstats/_runstats.py#L206

For the RMS rescale factor it might not matter too much, but when reporting RMSE values for forces I think this results in RMSE being ~40% (1-1/sqrt(3)) off.

### DIFF
--- a/mace/modules/utils.py
+++ b/mace/modules/utils.py
@@ -234,7 +234,7 @@ def compute_mean_rms_energy_forces(
     forces = torch.cat(forces_list, dim=0)  # {[total_n_graphs*n_atoms,3], }
 
     mean = to_numpy(torch.mean(atom_energies)).item()
-    rms = to_numpy(torch.sqrt(torch.mean(torch.square(forces)))).item()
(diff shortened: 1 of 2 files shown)


ilyes319:
Our implementation is actually the expected way we want to implement it. The mean squared error is defined by averaging across all dimensions, and the root mean squared error is just the root of that. The assumption is that every component of the predicted forces is a single, independent prediction of the model. As far I know, the vast majority of peoole follow that.

gabor1:
The distinction is force vector error or force component error. Both make sense. Most people report the latter (because it's smaller?)

misko:
Thanks for the response! Makes sense!

PR #392, opened 2024-04-23
run_train.py refers to an undefined "calc" when setting model_config["atomic_inter_shift"]

### LINKED ISSUE #380: undefined calc in multi-head-finetune branch run_train.py
In the `multi-head-finetune` branch, `mace/cli/run_train.py` line 432 
https://github.com/ACEsuit/mace/blob/f5ab9c5d45e09ac49c7874c44a14ba83c813d5ab/mace/cli/run_train.py#L432 refers to a variable `calc`, but there are many code paths (various nested ifs around https://github.com/ACEsuit/mace/blob/f5ab9c5d45e09ac49c7874c44a14ba83c813d5ab/mace/cli/run_train.py#L93) that don't define it

bernstei:
Check that `calc` exists before using it while setting model_config["atomic_inter_shift"]

closes #380

### DIFF
--- a/mace/cli/run_train.py
+++ b/mace/cli/run_train.py
@@ -86,6 +86,7 @@ def main() -> None:
     device = tools.init_device(args.device)
 
     if args.foundation_model is not None:
+        calc = None
         if args.foundation_model in ["small", "medium", "large"]:
             logging.info(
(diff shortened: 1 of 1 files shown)


ilyes319:
Fixed it in a more general way in commit 00cc869

PR #394, opened 2024-04-23
Allow MACECalculator to pick theory used for multihead models

### LINKED ISSUE #390: need to be able to specify what head is used for _prediction_ with a multihead model
Whether this is best done at the time of fitting (by masking out the pretrained head) or when creating the `MACECalculator`, there needs to be some way to clearly specify which head is used for prediction. The former is perhaps better for the use case of stabilizing fine tuning, but the latter is a bit more general

bernstei:
Initial form of this PR allows `MACECalculator` to override `Atoms.info["theory"]` when deciding which head to use for prediction. This is the minimal change I need to be able to use such models cleanly, but I'm not sure it's sufficient.

Other related things I've noticed
- if you don't specify a theory (i.e. nothing in info `"theory"` field), it prints an error about `'Default'` not matching. I think it's from https://github.com/ACEsuit/mace/blob/5a66ca442391a73e81bf4cadcfec22d6d31c3c96/mace/data/atomic_data.py#L131
- I'm not sure defaulting to any particular theory (first, as it does now, or any other arbitrary choice, like last) is a good idea to begin with. However, this choice might be related to details of the multi-head **fitting** (e.g. if we're using multi-head to stabilize fine-tuning, we could default to calling the new head a specific name, like `'Default'`).

closes #390

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -57,12 +57,14 @@ class MACECalculator(Calculator):
         charges_key="Qs",
         model_type="MACE",
         compile_mode=None,
+        multihead_theory=None,
         **kwargs,
     ):
         Calculator.__init__(self, **kwargs)
         self.results = {}
 
         self.model_type = model_type
+        self.multihead_theory = multihead_theory
 
         if model_type == "MACE":
             self.implemented_properties = [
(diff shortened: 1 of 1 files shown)


gabor1:
I suggest we rename "theory" to "head" or "multihead"

bernstei:
> I suggest we rename "theory" to "head" or "multihead"
 
I agree, and now, before a non-optimal name is ingrained

PR #423, opened 2024-05-15
removed wront argument parsing in preprocess_data

VondrakMar:
preprocess_data.py cannot be used with a tutorial provided code in the README file, because of the wrong parsing function


ilyes319:
good catch thank you

PR #427, opened 2024-05-16
Enable ASE calculator on intel.

wcwitt:
@ilyes319 @jharrymoore this is a little hacky but should enable use of the ASE calculator in Intel.

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -11,6 +11,8 @@ from typing import Union
 
 import numpy as np
 import torch
+# TODO: make this import conditional on intel somehow
+import intel_extension_for_pytorch as ipex
 from ase.calculators.calculator import Calculator, all_changes
 from ase.stress import full_3x3_to_voigt_6_stress
 
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -6,13 +6,18 @@
 
 
 from glob import glob
+from importlib.util import find_spec
 from pathlib import Path
 from typing import Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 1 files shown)


wcwitt:
I don't have time to figure out the conditional import at the moment (although it shouldn't be difficult) ... would you rather merge now or wait for that?

ilyes319:
@wcwitt Happy to merge now, but I am suprised that just this import does something. What is actually happening ?

RokasEl:
> I don't have time to figure out the conditional import at the moment (although it shouldn't be difficult) ... would you rather merge now or wait for that?

Just note that without a conditional import this will break MACE for users who don't have the `intel_extension_for_pytorch` package.

I think that at the very least the import should be surrounded by a `try ... except` block.

wcwitt:
> Just note that without a conditional import this will break MACE for users who don't have the intel_extension_for_pytorch package

Definitely true. It's possible other changes in this `intel` branch might have the same effect, which is why I wasn't too worried about it, but I'm not against doing things properly.

jharrymoore:
Previously I think I did the import only if the device was set to xpu.  Probably better is what Kelvin has done in the mace-ddp fork, something like:
```python
from importlib.util import find_spec
        
has_ipex = find_spec("intel_extension_for_pytorch")
if has_ipex:
    try:
        import intel_extension_for_pytorch as ipex
    except ImportError as e:
        raise ImportError("IPEX is installed, but could not be imported!") from e
```

ilyes319:
happy to merge if someone does that.

wcwitt:
I've just pushed something, testing now.

PR #443, opened 2024-06-04
get rid of all stress/n_atoms

bernstei:
get rid of `delta_stress_per_atom` and `RMSE_stress_per_atom` in tools/train.py, since stress is already intensive and does not need to be divided by number of atoms.


bernstei:
@ilyes319 what do you think of this one?

ilyes319:
Merging also.

PR #492, opened 2024-06-26
BUG: Single atoms not isolated incorrectly removed from training set

jameskermode:
One atom configurations which where not tagged with `config_type=IsolatedAtom` were incorrectly being filtered out of training sets. If I have understood correctly this is quite significant as it means all single-atom DFT configs (which often occur for bulk FCC and BCC structures, perhaps with random strains applied) were being omitted from training sets.

### DIFF
--- a/mace/data/utils.py
+++ b/mace/data/utils.py
@@ -242,21 +242,20 @@ def load_from_xyz(
         atoms_without_iso_atoms = []
 
         for idx, atoms in enumerate(atoms_list):
-            if len(atoms) == 1:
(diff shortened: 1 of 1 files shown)


jameskermode:
Note this bug doesn't occur in the `multi-head-inteface` branch, since there the check for length of Atoms being equal to one is moved inside the config type check: https://github.com/ACEsuit/mace/blob/multi-head-interface/mace/data/utils.py#L270-L272

ilyes319:
We corrected that at some point, but the change must have been reverted. Thank you for spotting that, I ll merge.

PR #535, opened 2024-07-27
Allow creation of MACECalculator without needing to write a checkpoint to disk.

CompRhys:
The current `MACECalculator` requires you to load from a particular structure of checkpoint file. If we use a different training framework our checkpoints are not compatible and so it would be nice to be able to just pass in models.

~~Still needs tests.~~

### DIFF
--- a/mace/calculators/foundations_models.py
+++ b/mace/calculators/foundations_models.py
@@ -227,4 +227,6 @@ def mace_anicc(
         print(
             "Using ANI couple cluster model for MACECalculator, see https://doi.org/10.1063/5.0155322"
         )
-    return MACECalculator(model_path, device=device, default_dtype="float64")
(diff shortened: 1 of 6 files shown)


CompRhys:
Sure, by checkpoint I mean a serialised file written to disk, by model I mean something like `ShiftScaleMACE` instantiated as an object, by training framework I mean something like pytorch lightning or accelerate. The checkpoint state dict for pytorch lightning cannot be loaded directly by this calculator instead requiring me to load the checkpoint access the model attribute save a compatible checkpoint to disk and then load the calculator with the path to that second compatible checkpoint. Resaving and reloading is unnecessary imo

CompRhys:
@ilyes319 is there any pathway to this being merged?

Issue #534, opened 2024-07-27
Move to ruff for linting and formating?


CompRhys:
Adopt ruff as a linter and formatter for the codebase as a more modern and faster alternative to black+pylint.

janosh:
fwiw, `ruff` has been running for months on `pymatgen`, `atomate2`, `chgnet` and others. it's been a great experience everywhere.

stenczelt:
Agreed, using `ruff` in professional codebases is very good as well, you can pick what you want to turn on from it, which we did gradually and alongside `pylint`, `flake8`, & `isort`  running in the CI still. 

`black`'s formatter was kept alongside though, partly due to better integration with IDEs and because it's not "fixing" everything else it has opinions/rules about, but only reformats the code. So you can have `black .` and `ruff check --fix` as separate actions, having the former run on save of files.

stenczelt:
> @CompRhys I know ruff is faster but is it less accurate than pylint? I had experiences where it was not as accurate.

You can get "inaccurate" behaviour from just turning it on on it's own and not turning on specific rules. I haven't dug deep into what differences there are in their rules and if `ruff` implements absolutely everything that `pylint` does.

CompRhys:
Should be closed following merge of #1319

PR #537, opened 2024-07-29
Ruff linter

CompRhys:
Addresses #534. Much faster linting and formatting using 1 tool as opposed to 3.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -1,28 +1,36 @@
-name: Pre-Commit Checks
+name: Linting
 
 on:
-  pull_request:
   push:
     branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
 
 jobs:
-  pre-commit:
+  lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
         with:
           python-version: "3.10"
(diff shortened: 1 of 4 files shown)


ilyes319:
Looking at your changes, some of them will break compatibility with Python version 3.7 and 3.8.

CompRhys:
I ran with `target-version = "py37"` and so I would assume that nothing would be broken. If you want to keep backwards compatibility to 3.7 would you take a PR adding a CI matrix as I shouldn't be installing an end-of-life python version without security fixes on work laptop?

ilyes319:
I think these kind of type hints: str | Path | None are not supported with older python versions. Why do you want to change the way we do type hints?

janosh:
> I think these kind of type hints: `str | Path | None` are not supported with older python versions.

they are actually, if you add `from __future__ import annotations` to the start of a module (inserting that import can be automated with `ruff`). it backports the leaner type annotations to 3.7

ilyes319:
Is there any guidelines from Python to prefere these kind of annotations?

CompRhys:
The pep for it was https://peps.python.org/pep-0604/ generally I think the preference has been to move towards not needing to explicitly import anything in order to type hint standard types such as list, str or combinations thereof.

janosh:
> Is there any guidelines from Python to prefer these kind of annotations?

yes, `from __future__ import annotations` is only necessary below 3.10. above, the new syntaxes works without side-effect imports which is part of python's move to enable more type safety

CompRhys:
closing as it seems that there's no interest from maintainers

PR #554, opened 2024-08-17
Create py.typed

Andrew-S-Rosen:
As recommended in [pep 0561](https://peps.python.org/pep-0561/), a blank `py.typed` marker should be included when type hints are used so downstream codes can type check with `mypy` and similar tools.


ilyes319:
ok with I will merge it.

Issue #596, opened 2024-09-18
mace_create_lammps_model fails to load mace_agnesi_small.model


stenczelt:
**Describe the bug**
Trying to convert a model to LAMMPS format seems to fail for some models.

**To Reproduce**

env setup
```
python3.11 -m venv venv
source ./venv/bin/activate
python -m pip install --upgrade pip
pip install mace-torch
```

download & run - failure
```
wget https://github.com/ACEsuit/mace-mp/releases/download/mace_mp_0b/mace_agnesi_small.model
mace_create_lammps_model mace_agnesi_small.model
```

results in
```
/private/tmp/mace-tmp/venv/lib/python3.11/site-packages/e3nn/o3/_wigner.py:10: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  _Jd, _W3j_flat, _W3j_indices = torch.load(os.path.join(os.path.dirname(__file__), 'constants.pt'))
/private/tmp/mace-tmp/venv/lib/python3.11/site-packages/mace/cli/create_lammps_model.py:13: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  model = torch.load(model_path)
Traceback (most recent call last):
[...]

PR #597, opened 2024-09-18
Fix for Issue 596

stenczelt:
Fix for #596, specifying `map_location` in `mace_create_lammps_model`


stenczelt:
Tested on MacOS (same system as in #596) and on a PyTorch 2.3 + CUDA 12.1 environment.

PR #634, opened 2024-10-11
Updating compiler support and test cases

hatemhelal:
torch compile is under active development so the mace compile support needed a few updates to keep up with recent torch releases.  This PR:
* adds pytest-benchmark to the development install of mace
* removes usage of an API that doesn't exist in pytorch 2.3 and up
* updates the test cases to use `ScaleShiftMACE` as I understand this is the most widely used model implementation
* removed the `fullgraph=True` option which was triggering a compiler assertion for using `requires_grad_` on the model input tensors. I think this assertion is ok to ignore since the test case that checks for graph breaks is still passing.
* skipping the mixed-precision benchmark as this fails with an assertion on mismatched datatypes in the scatter operation.

I'd prefer to leave the mixed-precision test case in place in case that can be made to work in some future PyTorch release.

### DIFF
--- a/mace/tools/compile.py
+++ b/mace/tools/compile.py
@@ -36,7 +36,7 @@ def prepare(func: ModuleFactory, allow_autograd: bool = True) -> ModuleFactory:
     """
     if allow_autograd:
         dynamo.allow_in_graph(autograd.grad)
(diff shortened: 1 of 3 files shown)


ilyes319:
Hey @hatemhelal, thank you very much. I had a look and it seems great.

PR #642, opened 2024-10-16
Fix multihead prediction for eval_configs.py

beckobert:
When constructing the `DataLoader`, it appears that the keyword passing the `heads` of the model to the function was missing, so I added it in the PR.

I also added an additional argument to the CL function, that should make it much more convenient to evaluate all structures on a single head. Specifying it for every structure in the extxyz file is a bit clunky, if that flexibility is not required.

Let me know, what you think or if I got something wrong and it was indeed working all along (it didn't for me).


ilyes319:
Hey @beckobert,
Looks great, thank you!

PR #663, opened 2024-10-29
Refactor data module

WillBaldwin0:
refactor the data module so that keys, properties and weights are not passed through many levels of function calls. 
Key change:
 - make a new class `KeySpecification` which holds two dictionaries, `arrays_keys` and `info_keys`. This object is created in `run_train` or (`preprocess_data`) and only this is passed through the data processing functions. Defaults are only set two places: Default keys are set in the `arg_parser` and default values are set only in the constructor for `AtomicData`.


RokasEl:
@ilyes319, I've already reviewed this PR you can see my and Will's chat here(https://github.com/WillBaldwin0/mace-1/pull/3).

This change makes the data module more extensible and reduced the number of arguments being passed to functions all over the code base. It also makes it easier to extend the AtomicData object by inheriting from it.

Finally, I've made a version of the `from_config` that I think is a bit cleaner (https://github.com/WillBaldwin0/mace-1/pull/4) but maybe you prefer the more explicit current version.

PR #675, opened 2024-11-06
Cluster force

LarsSchaaf:
Fitting to cluster forces for better intermolecular interactions. 

## Description 
Atoms are assumed to be separated into clusters, eg molecules. Each cluster has a unique id (which is passed to cluster_key). The cluster force is all forces belonging to a cluster summed together. The additional loss term fits the predicted to ground truth cluster force. 

## Possible extensions:
- Fit to torque

### DIFF
--- a/mace/data/atomic_data.py
+++ b/mace/data/atomic_data.py
@@ -36,6 +36,7 @@ class AtomicData(torch_geometric.data.Data):
     energy: torch.Tensor
     stress: torch.Tensor
     virials: torch.Tensor
+    cluster: torch.Tensor
     dipole: torch.Tensor
     charges: torch.Tensor
     weight: torch.Tensor
@@ -43,6 +44,7 @@ class AtomicData(torch_geometric.data.Data):
     forces_weight: torch.Tensor
     stress_weight: torch.Tensor
     virials_weight: torch.Tensor
+    cluster_weight: torch.Tensor
     dipole_weight: torch.Tensor
     charges_weight: torch.Tensor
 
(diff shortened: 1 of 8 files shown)


gabor1:
Is there a script to process an xyz to find clusters and annotate the file?

LarsSchaaf:
Will add annotation script from Ioan. 

### Another improvement to current implementation:
- at the moment the cluster ids have to be unique for the entire training/valid/test set. 
- could use double batch ID like Will does for kpoints 
- short term solution: check they are unique and throw error if not?

gabor1:
Make sure the script you add generates unique IDs, and then when reading the file you can check it. Why does the ID need to be unique across the data set rather than just within the config? You could generate a unique id by concatenating the config ID and the cluster ID and this would then be unique irrespective of the cluster ID uniqueness

LarsSchaaf:
Yes exactly, thats the to do ("double batched"). I'll edit it tomorrow.

ilyes319:
Could you add a small test for the loss + one in the run_train test?

LarsSchaaf:
The clustering scripts are taken from Ioan Magdau's repo: https://github.com/imagdau/aseMolec/blob/main/aseMolec/anaAtoms.py

ilyes319:
@LarsSchaaf I think it would be nice to add somme config_cluster_weight, inside the loss, as you already added them to the AtomicData.

```
  configs_weight = torch.repeat_interleave(
        ref.cluster_weight, ref.ptr[1:] - ref.ptr[:-1]
    ).unsqueeze(
        -1
    )
```
And then do a bunch of scatter etc to create a mask per cluster and mask out clusters that should not be weighted.

gabor1:
I don't know if it's currently possible or not, but it would be nice if not all atoms needed to be part of a cluster, just some.

LarsSchaaf:
At the moment it's fine for some configurations to not have any clusters. They just get given cluster IDs = 0.0. About assigning clusters only to part of the atoms, I think for many situations this will be roughly the same thing because the forces add to zero.

gabor1:
Can we please plot a way for this stuff to get into main

PR #679, opened 2024-11-09
use .4g as format gives better float representation for numbers in tr…

alinelena:
…aining, especially loss


alinelena:
i can force fixed width is that is needed but I really do not like the crazy 0.000000x, I am bad at counting.

gabor1:
well, you only need to count to 1 (compare with previous)

PR #721, opened 2024-11-28
now actually download the new models

alinelena:


mace/calculators/foundations_models.py
```diff
@@ -42,7 +42,7 @@ def download_mace_mp_checkpoint(model: Union[str, Path] = None) -> str:
 
     checkpoint_url = (
         urls.get(model, urls["medium"])
-        if model in (None, "small", "medium", "large")
+        if model in (None, "small", "medium", "large", "small", "small-0b", "medium-0b", "small-0b2", "medium-0b2", "large-0b2")
```
oerc0122, line 45:
```suggestion
        if model in (None, "small", "medium", "large", "small-0b", "medium-0b", "small-0b2", "medium-0b2", "large-0b2")
```

oerc0122: Do we want a top-level constant of these rather than one here repeated later?

ElliottKasoar: Or maybe `if model is None or model in urls`?

`urls` could be the top-level constant, to reuse it.

alinelena:
best will be high level constants

Issue #724, opened 2024-11-29
Unable to start finetuning


cecilia-hong:
Hello,

I hope you are all doing well, thank you for your help in my previous issue, unfortunately I have ran into another, this time when I tried to fine tune based on a foundation model. 

So for my first try, I followed the instructions here: https://mace-docs.readthedocs.io/en/latest/guide/finetuning.html 
 and my training input are as follows:

`mace_run_train \
    --name="MACE" \
    --foundation_model="small" \
    --multiheads_finetuning=False \
    --train_file="train.xyz" \
    --valid_fraction=0.05 \
    --test_file="test.xyz" \
    --energy_weight=1.0 \
    --forces_weight=1.0 \
    --E0s="average" \
    --energy_weight=100 \
    --forces_weight=1 \
    --lr=0.01 \
    --scaling="rms_forces_scaling" \
    --batch_size=2 \
    --max_num_epochs=6 \
    --ema \
    --ema_decay=0.99 \
    --amsgrad \
    --default_dtype="float64" \
    --device=cuda \
    --seed=3`

However, my training will not start and I cannot seem to understand the source of the error from the output files. I have attached my log file to this, can you please help me with this? 
[MACE_run-3_debug.log](https://github.com/user-attachments/files/17959946/MACE_run-3_debug.log)

ilyes319:
Does your node have internet access? Can you share your output log, there must be an error log somewhere.

cecilia-hong:
Hi, I see where the problem may be now as the computing nodes I use do not have internet access. Is internet access required even if I were to download the models/checkpoint from https://github.com/ACEsuit/mace-mp/releases/tag/mace_mp_0b to finetune as well?

Other than the log file I previously attached, the job output had the following:

`Traceback (most recent call last):
  File "/exports/csce/eddie/chem/groups/Hobday/Cecilia/anaconda3/envs/mace_new/bin/mace_run_train", line 8, in <module>
    sys.exit(main())
  File "/exports/csce/eddie/chem/groups/Hobday/Cecilia/anaconda3/envs/mace_new/lib/python3.10/site-packages/mace/cli/run_train.py", line 63, in main
    run(args)
  File "/exports/csce/eddie/chem/groups/Hobday/Cecilia/anaconda3/envs/mace_new/lib/python3.10/site-packages/mace/cli/run_train.py", line 126, in run
    calc = mace_mp(
[...]

PR #755, opened 2024-12-17
allow custom cache based on XDG_CACHE_HOME env variable, addresses #724

alinelena:


ilyes319:
@alinelena Thank you for that, should we perhaps make the env name more explicit or is there a reason you sticked to that name?

PR #793, opened 2025-01-21
changes to mace preprocess argparse to support configfile

jharrymoore:
update preprocess arg parser to accept a config file, in line with the main argparser


alinelena:
@jharrymoore i am under impression that something similar was already merged.... can you have a look and see if there is anything in this which is not already in

jharrymoore:
Yup I see this has done in from another PR, we can close this

PR #799, opened 2025-01-22
add python 3.13 to testing

alinelena:


alinelena:
@ilyes319 this is green again you may want it in sooner than later

PR #815, opened 2025-02-03
add omat model to foundations_models.py

ThomasWarford:
Can delete if too premature


ilyes319:
already done, thanks though!

PR #825, opened 2025-02-12
Revamped plot_train.py script

### LINKED ISSUE #524: Improve plotting script
**Feature request**
The plotting script currently lacks functionality: 
- Plotting cleaner different training loss/ validation loss 
- Add labels to the plot (swa, patience).

IsaacParker30:
Not sure if this is useful but I made some changes to the plot_train.py script. This maybe resolves #524

1. Ability to plot using log scale (base10) or linear scale on y axis (--linear arg).
2. Ability to select which force (based on ones printed in results file, such as rmse_f or mae_e_per_atom) (--energy_key and --force_key arg).
3. Ability to remove error / standard deviation region on plot as this could look ugly on log plot (--error_bars arg).
4. Ability to plot where second stage loss (swa) begins, using a dashed vertical line. (--swa_start arg)
5. Ability to change format of output file, for example png or pdf (--output_file arg)

I tried to keep all the ability to have the script plot the original way too. Think the labels may need some adjusting. Happy to do anything else needed if this is useful.

See pdfs and commands below on example (spikes) loss curve
Old way:
`python plot_train.py --PATH/TO/results/model.txt --output_format png --error_bars --min_epoch 50 --linear --force_key mae_f --energy_key mae_e`
[plot_train_old.pdf](https://github.com/user-attachments/files/18772602/plot_train_old.pdf)

_New_ way:
`python plot_train.py --PATH/TO/results/model.txt --output_format png --min_epoch 50 --swa_start 300`
[plot_train_new.pdf](https://github.com/user-attachments/files/18772601/plot_train_new.pdf)


gabor1:
plots are much improved! what about stress on the right hand panel?

gabor1:
I just had a chat with Will Baldwin, and we have two more suggestions. 

1) Add further panels below the current ones which shows the scatterplot of target vs model energies, force components and stress components, color coded for training set, validation set, and test set. 

2) call this plotting script automatically at the end of run_train.py . this will encourage everyone to look at it.

birdyLinch (dismissed):
LGTM

vue1999:
### plot_train.py (Standalone plotting script):
Updated with keys beyond energy and force, and added distinction between heads when multihead finetuning. 
plot_train
To run: 
```
python mace_folder_path/plot_train.py  --path paths_to_result.txt  --heads "default,pt_head" --keys "rmse_e,mae_f,rmse_stress"
```
The following quantities can be added as keywords:
```
    'mae_e', 'mae_e_per_atom', 'rmse_e', 'rmse_e_per_atom', 'q95_e',
    'mae_f', 'rel_mae_f', 'rmse_f', 'rel_rmse_f','q95_f',
    'mae_stress', 'rmse_stress',
    'rmse_virials_per_atom', 'mae_virials',
    'rmse_mu_per_atom'
```
### visualise\_train.py (Internal plotting: Generates real-time visualizations, including loss/RMSE curves and scatter plots of reported quantities.):
Any errors here won't stop the training.

Two arguments added:
--plot
--plot_frequency

By default, plots are generated only at the end of training ("--plot True", --plot_frequency 0). In this case scatter plots are evaluated for training, validation, and test data.

MACE_run-123_train_default_stage_two

To enable periodic updates during training, use "--plot_frequency N", where N specifies the interval in epochs (e.g., "--plot_frequency 5"). This will generate PNGs that update the loss/RMSE curves , along with scatter plots for training and validation data (**will not evaluate test set during training**).

To disable plotting entirely, set "--plot False".

Plots are done for the --error\_table quantities.
Note: Scatter plot evaluation requires inference on train/valid/test sets, which may introduce significant computational overhead.

ilyes319:
@vue1999  Perfect! Can you add a small section to the readme on "Plotting training" or something, for that functionality?

Issue #828, opened 2025-02-21
SLURM distributed


aschankler:
Distributed training requires `SLURM_NTASKS_PER_NODE` to be set, and if it is not, it provides a fairly unhelpful error message.

The variable is not set by slurm unless the `--ntasks-per-node` option is used, so there are some reasonable batch jobs (eg specifying `--ntasks` explicitly) where MACE fails unexpectedly. The variable is only used by MACE to compute a default if `SLURM_NTASKS` is not set, but based on my reading of the [SLURM doc](https://slurm.schedmd.com/sbatch.html#OPT_SLURM_NTASKS), `SLURM_NTASKS` will always be set if `NTASKS_PER_NODE` is set. Could the dependence on this environment variable be removed so that a wider range of scheduler requests can be accepted without error?

aschankler:
I, at least, would prefer a couple clearly documented environment variables or CLI options to the current situation. However some of the other parameters (eg job rank) could be a little more difficult to to get manually.

bernstei:
I personally really don't like the code being scheduler specific (comes from having used many schedulers over the years).  I especially hate it when it's slurm, which is in general nice but terrible about not setting env vars.

alinelena:
there is also this one https://github.com/ACEsuit/mace/pull/586 situation at this stage is suboptimal. to its sins slurm IF set correctly just works... personally I would suggest having pre-canned options and one for total freedom...

ilyes319:
Long time after but that was fixed in recent releases:

https://github.com/ACEsuit/mace/blob/42cd495b9c8e155441c405fef842a8ba5107c2f8/mace/tools/slurm_distributed.py#L26
and we now support more distributed envs:
https://github.com/ACEsuit/mace/blob/main/mace/tools/distributed_tools.py

make sure to reopen if that's not good enough.

PR #831, opened 2025-02-24
Allow other schedulers for distributed training

aschankler:
Add an argument to disable automatically inferring distributed parameters from the SLURM scheduler, as discussed in #828. The default behavior should remain unchanged


ilyes319:
Long time after but that was fixed in recent releases:
https://github.com/ACEsuit/mace/blob/42cd495b9c8e155441c405fef842a8ba5107c2f8/mace/tools/slurm_distributed.py#L26
make sure to reopen if that's not good enough.

PR #836, opened 2025-02-27
Add functionality to select heads to evaluate error tables on and perform dry runs.

IsaacParker30:
Added 2 new arguments:

`--eval_heads` 
This allows the user to specify which heads they want to evaluate / print an error table for at the end of training. If not set, will default evaluate all heads.
Usage: Specify all heads you want to evaluate on as a string with heads separated by commas. For example:
`--eval_heads='default,pt_head,head3'`
Example use scenario would be for replay fine-tuning if the user doesn't want to evaluate the large foundation model database on the `pt_head`.

`--dry_run` 
Adding this argument will stop the `run_train.py` script just before the model training is about to begin (calls the `tools.train()` function. This allows the user to check if they've set their parameters correctly before beginning an expensive training run.

I have tested:

- If dry run works.
- If I can select heads to turn off.
- If code evaluates on all heads if none specified.
- If code stops if user specifies a head to evaluate on that doesn't exist.

Let me know if any changes needed.

### DIFF
--- a/mace/cli/run_train.py
+++ b/mace/cli/run_train.py
@@ -423,6 +423,18 @@ def run(args) -> None:
             logging.debug(
                 "Using LMDB/HDF5 datasets for pretraining or fine-tuning - skipping ratio check"
             )
+    
+    # Find heads to evaluate in error table
+    if args.eval_heads is None:
+        eval_heads = heads
+    else:
+        eval_heads = list(args.eval_heads.split(","))
+        for eval_head in eval_heads:
+            if eval_head not in heads:
+                logging.error(f"Head {eval_head} not found in the list of heads: {heads}.")
+                raise KeyError(f"Head {eval_head} not found in the list of heads: {heads}.")
+
+    logging.info(f"Will evalute error table on heads: {eval_heads}")
 
     # Atomic number table
     # yapf: disable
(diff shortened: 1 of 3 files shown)


mace/tools/tables_utils.py
```diff
@@ -100,6 +101,9 @@ def create_error_table(
         ]
 
     for name in sorted(all_data_loaders, key=custom_key):
+        #if name not in eval_heads: 
+        #    logging.debug(f"Not evaluating {name} as not user request. SKIP")
+        #    continue
```
vue1999, line 106:
```suggestion
```

vue1999, line 106: Since train_valid and test data are already filtered based on eval_heads, this step isn't needed. We can remove the eval_heads input and the List import too then.

vue1999 (changes requested):
Please fix the pre-commit check suggestions, otherwise looks good.

mace/cli/run_train.py
```diff
@@ -352,6 +352,21 @@ def run(args: argparse.Namespace) -> None:
...
+    else:
+        eval_heads = list(args.eval_heads.split(","))
+        for eval_head in eval_heads:
+            if eval_head not in heads:
+                logging.error(
+                    f"Head '{eval_head}' not found in the list of heads: {heads}."
+                )
+                raise ValueError(
+                    f"Head '{eval_head}' not found in the list of heads: {heads}."
+                )
```
vue1999, line 367: Collect all eval heads not in heads and raise a single error to avoid multiple runs for multiple misdefined eval heads

vue1999 (dismissed):

alinelena:
@IsaacParker30 this addresses some of the things in #839 should have checked the PR before writing it. I wonder if makes sense to implement also first part in this.

alinelena:
one small point, if you do not need the head stats at the end you may not want to print it during the training either.

alinelena:
@IsaacParker30 this will need a small rebase

alinelena:
@IsaacParker30 any news on this?

IsaacParker30:
@alinelena I have rebased (I think...). After discussion leading from #839 I have also removed the printing error tables during training functionality.

alinelena:
ok, I still see no way for a normal user to determine if valid/test split is good without keeping an eye on that errors... can we at least save the intermediate models... maybe this can be done when plotting? @vue1999

ilyes319:
@IsaacParker30 Thank you very much for that, I accidentally redone that (expect i prefer to filter heads instead of selecting them). I made a co commit to add the dry run too: https://github.com/ACEsuit/mace/tree/e3f4c4b96d349e0716ebd218725f653597f2250b

PR #845, opened 2025-03-06
Updating calculator weights from checkpoints

LarsSchaaf:
If you want to see performance changes throughout epochs, you can load a single model and then update the weights from former checkpoints. 

## To do: 
- create test: where you compare the output from loading two separate models or updating based on their corresponding checkpoints.

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -370,6 +370,37 @@ class MACECalculator(Calculator):
                     .numpy()
                 )
 
+    def update_weights_from_checkpoint(self, checkpoint_path):
+        """
+        Updates the model's weights from a given checkpoint file.
+
+        Args:
+            checkpoint_path (str): Path to the checkpoint file.
+
+        Example:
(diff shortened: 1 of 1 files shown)


RokasEl:
What's the use case of doing:
```python
calc.update_weights_from_checkpoint(LOAD_PATH)
```
compared to:
```python
calc = MACECalculator(model_paths=LOAD_PATH)
```
?

If there is a genuine case, I think it would be better to modularize the `__init__` of the calculator to use a `load_weights` function and use that for both initializing the calculator, and updating the weights inplace.

LarsSchaaf:
The difference is that one loads `.model` s and this one can load checkpoints `.pt` files, ie just state_dicts

LarsSchaaf:
### Note: 
- `state_dict` changes when using [cuEquivariance](https://github.com/NVIDIA/cuEquivariance), so can only update models with checkpoints with same cuequivariance setting
- compiled models also don't work atm

PR #854, opened 2025-03-10
Add a BatchRelaxer to speed up geometry optimisation of many small structures

zhubonan:
Add a batch relaxer which combines multiple atoms as a single batch to the underlying model. This can give speed-ups when lots of smaller structures needs to be relaxed using GPU (typical while using the foundation models for high-throughput studies).  

The current code is working, but a few useful functionalities can still be added:
- [ ] Support for ensemble models
- [ ] Support other types of MACE like `EnergyDipoleMACE`?
- [ ] Since we already run the optimiser step-by-step, it does no harm to add some detection for pathological structures (exploded/collapsed) which can make the main worker process to stuck.
- [ ] Automatically determine the best batch size by collecting the timings of the `compute` method.

### DIFF
--- a/mace/calculators/batch_relax.py
+++ b/mace/calculators/batch_relax.py
@@ -2,7 +2,6 @@
 Module for doing batch relaxation
 """
 from typing import List, Dict, Any
-import torch
 from ase import Atoms
 from ase.calculators.singlepoint import SinglePointCalculator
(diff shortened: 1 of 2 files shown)


ilyes319:
Hey looks great, ping me when you think it is ready to merge.

PR #871, opened 2025-03-16
Making MACE using cuEquivariance work with Torch Script

cbatton:
I have an application where I need to use TorchScript, and noticed that it wasn't working with cuEquivariance in the latest version. I have isolated it down to the two instances in the code where a "self.cueq_config" variable is set, with TorchScript not being compatible with the CuEquivarianceConfig class. I have replaced the "self.cueq_config" values with flags taken from cueq_config. A small working example is attached that is runnable with "python test.py --xyz_file=carbon.xyz", which should fail in the current version of MACE but work with the proposed changes.

[test_mace_script.tar.gz](https://github.com/user-attachments/files/19269280/test_mace_script.tar.gz)

### DIFF
--- a/mace/modules/blocks.py
+++ b/mace/modules/blocks.py
@@ -275,7 +275,16 @@ class EquivariantProductBasisBlock(torch.nn.Module):
             shared_weights=True,
             cueq_config=cueq_config,
         )
-        self.cueq_config = cueq_config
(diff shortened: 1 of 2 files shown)


ilyes319:
Hello, cueq should not yet be compatible with torchscript for other internal reasons but it is coming soon.

PR #876, opened 2025-03-17
adding XPU to training, evaluation, and inference w/ ASE 

alvarovm:
I noticed that while training and performing inference using XPU devices, offloading was missing in some parts. Additionally, this was not explicitly handled in the ASE's MACECalculator.


ilyes319:
I have already merged this manually, so closing now. Thank you very much! You can use on the develop branch.

PR #879, opened 2025-03-20
Refactor data update

RokasEl:
# Release Note: Improved Data Handling with KeySpecification

This release introduces the `mace.data.KeySpecification` class to streamline and clarify how property keys (like energy and forces) are handled within MACE, improving code maintainability and reducing potential confusion. This merge is also necessary to enable a future merge of electrostatics support.

**Key Changes:**

* **Centralized Key Definitions:** The default keys for properties are now defined in a single, easily accessible `DefaultKeys` Enum. This eliminates redundant definitions throughout the codebase and ensures consistency.

* **Introduction of `KeySpecification`:** The `config_from_atoms` function now utilizes the `KeySpecification` class to manage property keys.

    **Before:**
    ```python
    config = config_from_atoms(atoms, energy_key=..., forces_key=...,)
    ```

    **After:**
    ```python
    config = config_from_atoms(atoms: ase.Atoms, key_specification: KeySpecification = KeySpecification(),)
    ```
    The default `KeySpecification` is empty. To use default keys, construct it like this:
    ```python
    key_spec = KeySpecification.from_defaults()
    ```

* **Explicit Head Choice for MACECalculators Created from Multihead Trained Models**  For models trained with multiple heads, the `MACECalculator` now requires you to explicitly specify the head to use if the first head is not named "Default" or "default". This prevents unexpected behavior and makes head selection more transparent.

* **Change of Default Settings for Finetuning `MP` Models** Previously, the default settings for selecting data for replay finetuning were such that MPTraj data were first filtered to only contain the same combinations as those in your training data then downselected to `num_samples_pt`. Reading all of the training data just to determine the unique elements can be quite slow especially with large datasets, as such, the new default behaviour is to just use a random subset of the MPTraj data.

**Impact:** If you are loading a multihead model and encounter a `ValueError`, you will need to specify the desired head when creating the `MACECalculator`.

**Benefits:**

* **Increased Clarity:**  Property key definitions are now centralized and easier to find.
* **Improved Consistency:**  Ensures consistent default keys across the codebase.
* **Reduced Redundancy:** Eliminates repeated definitions, simplifying maintenance.
* **More Explicit Behavior:** The change in `MACECalculator` for multihead models makes the behavior more predictable and less prone to silent errors.

**Migration Notes:**

*  If you were relying on the implicit default keys in `config_from_atoms`, ensure you now use `KeySpecification.from_defaults()` or explicitly provide your desired `KeySpecification`.
*  When loading multihead models with `MACECalculator`, double-check the head names and specify the desired head if necessary.
* As noted above, this PR changes the default choice of data for replay finetuning. If you want to keep the old settings, you need to specify the elements to keep in the replay data directly, e.g. `--atomic_numbers "[1, 8]" and add this flag `--filter_type_pt combinations`.


alinelena:
one thing I noticed, in ft, evaluating pt_head when num samples is 100000 is terrible slow...., can you check if you inherited that in your refactoring?

RokasEl:
> one thing I noticed, in ft, evaluating pt_head when num samples is 100000 is terrible slow...., can you check if you inherited that in your refactoring?

I think this is now the recommended way to do finetuning. That is, each epoch is slower but you need *less* epochs overall. Maybe @ilyes319 can weigh in as well.

alinelena:
@RokasEl nope is not that after the last data refactoring (lmdb added) evaluating the train_pt_head after the ft become crazy slow. for example will take me 2h to ft and 5h to compute the tables at the end... for 100K samples

PR #893, opened 2025-03-31
eval_configs: calculate descriptors + cueq

ThomasWarford:


alinelena:
if you go this direction... I would suggest to consider... per atom and per species descriptors to be returned

ThomasWarford:
If I understand you correctly that is what I have done. The user can set `--descriptor_aggregation_method` to `mean`, `per_element_mean` or not specify it to get descriptors for each atom.

mace/cli/eval_configs.py
```diff
@@ -49,6 +57,30 @@ def parse_args() -> argparse.Namespace:
...
+    parser.add_argument(
+        "--descriptor_num_layers",
+        help="number of layers to take descriptors from",
+        type=int,
+        default=-1,
+    )
+    parser.add_argument(
+        "--descriptor_aggregation_method",
+        help="method for aggregating node features. None saves descriptors for each atom.",
+        choices=["mean", "per_element_mean",],
```
alinelena, line 75: I would also generate per_atom, may be useful some times.

ThomasWarford: None (per atom) is the default, but I'll make this more explicit

ilyes319:
@ThomasWarford is that ready?

ThomasWarford:
> @ThomasWarford is that ready?

Yes

PR #896, opened 2025-04-01
fix swa stress logging

vue1999:
Fixes second stage weight logging mistake, that printed first stage stress weight instead of second stage. Raised in https://github.com/ACEsuit/mace/discussions/895


ilyes319:
good catch!

PR #922, opened 2025-04-16
Virials log

vue1999:
When training on virials, the logger still printed stress error metrics.
Although special-case conditions were defined for virial-specific logging, they were never triggered — because the stress metric was always created with zero values, causing the logger to fall back to the stress error path.


ilyes319:
For mixed data and torchscript reasons, it is impossible to return None for the stresses unfortunately, that is why i removed it.

ilyes319:
We need to find another solution to the logging problem, but that will probably be based on a whole refont of the logger.

PR #926, opened 2025-04-20
add matpes models + fix the sign for edge forces

ilyes319:


bernstei:
Does this PR allow for multihead fine-tuning of these models, or is there no pretrained data available for that?

ilyes319:
> Does this PR allow for multihead fine-tuning of these matpes models, or is there no pretrained data available for that?

That's high on my TODO, not yet included.

PR #931, opened 2025-04-24
MACECalculator search for head named 'default' when no head is specified explicitly

### LINKED ISSUE #930: current mace can't read previously generated multihead fine tuned models because of head name or ordering
I'm not sure what the preferred solution is going to be, but we have users here who generated models with multihead fine tuning a couple of months ago, and the resulting models have a head named `"default"`, but now the default loading code looks for `"Default"` which it can't find.

```
  File "/opt/local/python/wif/conda/bin/wif_md", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/local/python/wif/conda/lib/python3.11/site-packages/wif/cli/md.py", line 161, in main
    md(parser.parse_args(*cli_args))
  File "/opt/local/python/wif/conda/lib/python3.11/site-packages/wif/cli/md.py", line 57, in md
    md_calc = md_calc[0](*md_calc[1], **md_calc[2])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/calculators/mace.py", line 211, in __init__
    raise ValueError(
ValueError: Head keyword was not provided, and the head in the model is not 'Default'Please provide a head keyword to specify the head you want to use. Available heads are: ['pt_head', 'default']
```

Should we change the search for heads to be non-case-sensitive?  Do I need to convert all the models, and if so, does `mace_select_head` do that?

bernstei:
When defaulting to a head named default from a multihead finetuned model, search, don't just assume it's the first in available list.

closes #930

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -206,12 +206,14 @@ class MACECalculator(Calculator):
         if kwarg_head is not None:
             self.head = kwarg_head
         else:
-            self.head = self.available_heads[0]
-        if kwarg_head is None and self.head.lower() != "default":
-            raise ValueError(
-                "Head keyword was not provided, and the head in the model is not 'Default'"
-                f"Please provide a head keyword to specify the head you want to use. Available heads are: {self.available_heads}"
-            )
+            self.head = [head for head in self.available_heads if head.lower() == "default"]
+            if len(self.head) == 0:
+                raise ValueError(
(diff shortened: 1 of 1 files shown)


RokasEl (approved):
This is definitely a better default behaviour. We should merge this asap.

PR #932, opened 2025-04-25
add config for options  fine tuning selections

alinelena:
this just adds --config to mace_finetunning (which selects samples despite the name).

add --atomic_numbers option

rename it to something more sensible?

### DIFF
--- a/mace/calculators/foundations_models.py
+++ b/mace/calculators/foundations_models.py
@@ -32,8 +32,7 @@ mace_mp_urls = {
     "mace-matpes-pbe-0": "https://github.com/ACEsuit/mace-foundations/releases/download/mace_matpes_0/MACE-matpes-pbe-omat-ft.model",
     "mace-matpes-r2scan-0": "https://github.com/ACEsuit/mace-foundations/releases/download/mace_matpes_0/MACE-matpes-r2scan-omat-ft.model",
 }
-mace_mp_constants = list(mace_mp_urls.keys())
-mace_mp_names = [None] + mace_mp_constants
+mace_mp_names = [None] + list(mace_mp_urls.keys())
 
 
 def download_mace_mp_checkpoint(model: Optional[Union[str, Path]] = None) -> str:
--- a/mace/cli/fine_tuning_select.py
+++ b/mace/cli/fine_tuning_select.py
@@ -18,7 +18,7 @@ import torch
 from tqdm import tqdm
 
 from mace.calculators import MACECalculator, mace_mp
-from mace.calculators.foundations_models import mace_mp_constants
+from mace.calculators.foundations_models import mace_mp_names
 
 try:
     import fpsample  # type: ignore
@@ -281,7 +281,6 @@ class FPS:
 
         for i, atoms in tqdm(
             enumerate(self.atoms_list),
-            total=len(self.atoms_list),
             desc="Assembling descriptors",
         ):
             descriptors = atoms.info["mace_descriptors"]
@@ -296,7 +295,7 @@ def _load_calc(
 ) -> Union[MACECalculator, None]:
     if subselect == SubselectType.RANDOM:
         return None
-    if model in mace_mp_constants:
+    if model in filter(None, mace_mp_names):
         calc = mace_mp(model, device=device, head=head, default_dtype=default_dtype)
     else:
         calc = MACECalculator(
@@ -316,7 +315,7 @@ def _get_finetuning_elements(
             "Using elements from the finetuning configurations for filtering."
         )
         species = np.unique([x.symbol for atoms in atoms for x in atoms]).tolist()  # type: ignore
-    elif filter_atomic_numbers_pt is not None and filter_atomic_numbers_pt:
+    elif filter_atomic_numbers_pt:
         logging.debug("Using the supplied atomic numbers for filtering.")
         species = [ase.data.chemical_symbols[z] for z in filter_atomic_numbers_pt]
     else:
@@ -531,7 +530,7 @@ def select_samples(
             "suffix compatible with extxyz format"
         )
     output_path = Path(settings.configs_pt).name
-    _maybe_save_descriptors(filtered_pt_atoms, output_path, False)
+    _maybe_save_descriptors(filtered_pt_atoms, output_path, True)
     _maybe_save_descriptors(subsampled_atoms, settings.output)
 
     _write_metadata(


mace/cli/fine_tuning_select.py
```diff
@@ -136,12 +159,13 @@ def parse_args() -> argparse.Namespace:
...
     parser.add_argument("--seed", help="random seed", type=int, default=42)
-    return parser.parse_args()
+    parser.add_argument("--atomic_numbers", help="atomic numbers to keep for filtering", nargs="+", type=int, required=False)
+    return parser
 
 
 def calculate_descriptors(atoms: List[ase.Atoms], calc: MACECalculator) -> None:
     logging.info("Calculating descriptors")
-    for mol in atoms:
+    for mol in tqdm(atoms,total=len(atoms), desc="Calculate descriptors"):
```
bernstei, line 168: doesn't tqdm default to `len(iterable)` ?

mace/cli/fine_tuning_select.py
```diff
@@ -136,12 +159,13 @@ def parse_args() -> argparse.Namespace:
         default=1.0,
     )
     parser.add_argument("--seed", help="random seed", type=int, default=42)
-    return parser.parse_args()
+    parser.add_argument("--atomic_numbers", help="atomic numbers to keep for filtering", nargs="+", type=int, required=False)
```
bernstei, line 162: Here this seems like a sensible name for the option, but if it's also going to be in `mace_run_train`, _there_ it needs a more explicit name.  Maybe `--pt_filter_atomic_numbers` ?

bernstei:
Do you want to merge this with #965 ?

mace/cli/fine_tuning_select.py
alinelena: you are right I just like to be explicit

alinelena:
> Do you want to merge this with #965 ?

yes do you want to merge it in? I suspect you can PR it to my fork/branch

mace/cli/fine_tuning_select.py
alinelena: @ilyes319 any thoughts on name?

alinelena:
> Do you want to merge this with #965 ?

@ilyes319 shall we merge that one before?

mace/cli/fine_tuning_select.py
```diff
@@ -55,17 +63,33 @@ class SelectionSettings:
...
 def str_to_list(s: str) -> List[int]:
     assert isinstance(s, str), "Input must be a string"
     return ast.literal_eval(s)
 
 
-def parse_args() -> argparse.Namespace:
-    parser = argparse.ArgumentParser(
-        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
-    )
+def build_default_finetuning_select_arg_parser() -> argparse.ArgumentParser:
```
bernstei, line 74: I have a suggestion that would help merge this with doing filtering in `mace_run_train`.  Have this function get an optional parser argument, and a suffix.  If the parser argument is passed, the options (I guess not `--config`) are added to that one, otherwise it creates a new parser object.  It also adds the suffix to all the arguments, so that `mace_run_train` can pass `suffix=_mp` and things like `--filter_type`, which makes sense for `mace_finetune_select`, will become `--filter_type_mp`, the way they are now in the `develop` branch.

setup.cfg
```diff
@@ -43,7 +43,7 @@ console_scripts =
     mace_plot_train  =  mace.cli.plot_train:main
     mace_run_train  =  mace.cli.run_train:main
     mace_prepare_data  =  mace.cli.preprocess_data:main
-    mace_finetuning = mace.cli.fine_tuning_select:main
+    mace_finetuning_select = mace.cli.fine_tuning_select:main
```
bernstei, line 46: I strongly agree!

mace/cli/fine_tuning_select.py
```diff
@@ -30,18 +32,24 @@ class FilteringType(Enum):
...
+        return self.value
+
 
 @dataclass
 class SelectionSettings:
     configs_pt: str
     output: str
     configs_ft: str | None = None
-    atomic_numbers: List[int] | None = None
+    pt_filter_atomic_numbers: List[int] | None = None
```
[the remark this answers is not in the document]
alinelena: @bernstei because you suggested in a previous comment.

ilyes319:
@alinelena @bernstei Can you both remind me what's the target of that PR, and how that fits in the latest Noam's PRs pipeline?

mace/cli/fine_tuning_select.py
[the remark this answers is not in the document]
bernstei: Oh, wait, it's more complicated than I thought. This code can process both fine-tuning and pretraining, although I never use it that way.  Is there really a reason for it to deal with the ft configs?  In any case, since it's `configs_pt` and `configs_ft`, if we want to add that, it should be `filter_atomic_numbers_pt`, probably.

But if we want to make the interface consistent with `mace_run_train`, lots of argument names would have to be rethought, and my idea for making them consistent by having the same code add arguments to both parser will also have to be rethought.

bernstei:
I'm not planning to use this (just filter from inside `mace_run_train`), although I think it wouldn't be a bad idea for this to exist as a standalone command line interface that does _the same data filtering and processing_ as `mace_run_train`.  

But I think that if there are both, it'd be good to make those command line arguments as consistent as possible.

mace/cli/fine_tuning_select.py
bernstei: I just noticed, above, that this won't work right now because this code also handles fitting data at the same time. I'd like to suggest going through all the command line arguments and making them as much like `mace_run_train` as possible.

alinelena:
@bernstei when you filter from mace_run_train, can be quite expensive, especially if you want or have to do the descriptors too. I will rebase and try to get it done

mace/calculators/foundations_models.py
```diff
@@ -32,7 +32,8 @@
     "mace-matpes-pbe-0": "https://github.com/ACEsuit/mace-foundations/releases/download/mace_matpes_0/MACE-matpes-pbe-omat-ft.model",
     "mace-matpes-r2scan-0": "https://github.com/ACEsuit/mace-foundations/releases/download/mace_matpes_0/MACE-matpes-r2scan-omat-ft.model",
 }
-mace_mp_names = [None] + list(mace_mp_urls.keys())
+mace_mp_constants = list(mace_mp_urls.keys())
```
bernstei: What is "constants" supposed to convey here?  And what is mace_mp_names actually used for?

mace/cli/fine_tuning_select.py
```diff
@@ -263,32 +292,33 @@ def assemble_descriptors(self) -> None:
...
-    atoms: List[ase.Atoms], atomic_numbers: List[int] | None
+    atoms: List[ase.Atoms], filter_atomic_numbers_pt: List[int] | None
 ) -> List[str]:
     if atoms:
         logging.debug(
             "Using elements from the finetuning configurations for filtering."
         )
         species = np.unique([x.symbol for atoms in atoms for x in atoms]).tolist()  # type: ignore
-    elif atomic_numbers is not None and atomic_numbers:
+    elif filter_atomic_numbers_pt is not None and filter_atomic_numbers_pt:
```
bernstei: test for not None is superfluous - None evaluates to false

bernstei:
Other than the few comments I put in, it looks fine, except there are random spacing changes that lead to more diffs than are actually relevant.

mace/cli/fine_tuning_select.py
```diff
@@ -254,7 +279,11 @@ def assemble_descriptors(self) -> None:
             dtype=np.float32,
         ).astype(np.float32)
 
-        for i, atoms in enumerate(self.atoms_list):
+        for i, atoms in tqdm(
+            enumerate(self.atoms_list),
+            total=len(self.atoms_list),
```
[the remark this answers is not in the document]
alinelena: removed

mace/calculators/foundations_models.py
alinelena: ok was a reason for it but i fixed it using filter None since was used in only one place.

bernstei:
Looks mostly good now, except some random spacing changes in `tests/test_run_train_dipole_polar.py` and `tests/test_run_train.py` that aren't meaningful.  

Also, is there a test for this CLI script?

PR #937, opened 2025-04-29
fix keyspec being modified inplace

RokasEl:
Fixes the keyspec object being modified in place during multihead finetuning.

Also adds tests that check for this behaviour.

This fixes issue 933.

### DIFF
--- a/mace/cli/run_train.py
+++ b/mace/cli/run_train.py
@@ -215,7 +215,7 @@ def run(args) -> None:
         pt_keyspec = (
             args.heads["pt_head"]["key_specification"]
             if "pt_head" in args.heads
-            else args.key_specification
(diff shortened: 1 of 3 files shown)


RokasEl:
Note, test failed due to error in model download. Likely an internal github issue.

bernstei:
Is it clear how `tests/test_run_train.py::test_run_train_foundation` passed with before this bug was fixed? Or was it actually failing, and merged anyway?

ilyes319:
@RokasEl Weird that the tests fail, probably a temporary github problem.
@bernstei I don't get how all the tests pass and this basis functionality does not work, especially because i have doing the same thing many times; probably has to do with the way i renamed the keys.

PR #944, opened 2025-05-03
OpenEquivariance Integration

vbharadwaj-bk:
Adds support for OpenEquivariance in MACE (in particular, unlocks acceleration on AMD architectures). Integration follows a similar pattern to CUEQ. Adds options for a fused TP + scattersum with atomics or the unfused operation, for now.

Major diffs:
1. Scatter sum moved to wrapper_ops and combined with conv_tp. With this change of interface (which should not affect model loading from a saved state), we gain the ability to fuse the convolution with the TP and achieve significant memory savings.
2. test_cue.py renamed to test_cue_oeq.py and runs the roundtrip comparisons for both kernel providers.

This is a prototype, and we are open to any changes and will adapt to your needs. We do not provide a CPU backend for our kernels, in particular, but we could provide a fallback if this is needed.

@asglover

### DIFF
--- a/mace/cli/convert_e3nn_oeq.py
+++ b/mace/cli/convert_e3nn_oeq.py
@@ -32,7 +32,7 @@ def run(
     config["oeq_config"] = OEQConfig(
         enabled=False,
         optimize_all=True,
-        conv_fusion=None
+        conv_fusion="atomic"
     )
 
     # Create new model with oeq config
--- a/mace/cli/convert_e3nn_oeq.py
+++ b/mace/cli/convert_e3nn_oeq.py
@@ -1,0 +1,89 @@
+import argparse
+import logging
+import os
+from typing import Dict, List, Tuple
+
+import torch
+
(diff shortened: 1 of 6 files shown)


ilyes319:
@vbharadwaj-bk Thank you very much for the PR! Looking forward to merging it. Could you have a look at the different PR errors, hopefully it should be quite easy to solve them. Thanks a lot!

vbharadwaj-bk:
Linted and fixed!

ilyes319:
thank you! @vbharadwaj-bk there still seems to be a small problem with the import in test. Can you make sure we can add oeq as an optional dependency?

vbharadwaj-bk:
Oops - these are fixable, I'll take a look when Perlmutter is back up.

vbharadwaj-bk:
Fixed, apologies this took a bit longer than expected. I also ran the tests on a Perlmutter CPU node (as well as the CUEQ / OEQ tests on a GPU node) to verify that they all pass.

PR #965, opened 2025-05-14
Use pathlib to better handle suffixes in finetuning_select

### LINKED ISSUE #964: fine_tuning_select overwrites files because it assumes .xyz as the file suffix
This kind of statement (also other places in the same file)
https://github.com/ACEsuit/mace/blob/b5faaa076c49778fc17493edfecebcabeb960155/mace/cli/fine_tuning_select.py#L481
need to be redone so they don't assume particular suffixes (in particular .xyz vs .extxyz),  because they overwrite if the assumptions are wrong.

Pretty easy to do in a fairly general way with pathlib `Path.stem` and `Path.suffix`

bernstei:
closes #964

### DIFF
--- a/mace/cli/fine_tuning_select.py
+++ b/mace/cli/fine_tuning_select.py
@@ -8,6 +8,7 @@ import logging
 from dataclasses import dataclass
 from enum import Enum
 from typing import List, Tuple, Union
+from pathlib import Path
 
 import ase.data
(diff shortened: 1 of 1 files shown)


bernstei:
@ilyes319 any chance you can review this so we can stop overwriting perfectly valid files?

PR #975, opened 2025-05-17
Pseudolabel finetuning

vue1999:
Add an option to use the pseudolabels generated with the foundation model when using multihead finetuning with replay


alinelena:
@vue1999 can you expand on documentation?

PR #1006, opened 2025-06-07
Fix oeq import handling on macos without cuda

CompRhys:
Various tests fail when oeq installed but cuda not available.

See https://github.com/PASSIONLab/OpenEquivariance/issues/121


vbharadwaj-bk:
Problem on our end, sorry about this - corrected with https://github.com/PASSIONLab/OpenEquivariance/pull/122, which raises an ImportError (so that the guard catches it) when the extension library fails to load.

Issue #877, opened 2025-03-19
mixing creation and use of float32 and float64 calculators leads to a crash


bernstei:
I suppose this might be too weird of an edge case to fix, but the following code 

```
from ase.atoms import Atoms

from mace.calculators import mace_mp

atoms = Atoms('H', cell=[2] * 3, pbc=[True] * 3)

calc64 = mace_mp("small", device="cpu", default_dtype="float64")
calc32 = mace_mp("small", device="cpu")

calc32.get_descriptors(atoms=atoms)

calc64.get_descriptors(atoms=atoms)
```

results in the following output and error. I'm guessing that something is getting cached module-wide, and that ends up with mixed type in the error message

```
/home/cluster/bernstei/.local/lib/python3.11/site-packages/e3nn/o3/_wigner.py:10: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  _Jd, _W3j_flat, _W3j_indices = torch.load(os.path.join(os.path.dirname(__file__), 'constants.pt'))
Using Materials Project MACE for MACECalculator with /home/cluster/bernstei/.cache/mace/20231210mace128L0_energy_epoch249model
Using float64 for MACECalculator, which is slower but more accurate. Recommended for geometry optimization.
[...]

Issue #328, opened 2024-02-24
Don't modify torch default dtype


janosh:
this line prevents running other models after MACE for relaxation in the same Python session since MACE recommends float64 for geometry optimization while e.g. chgnet and m3gnet use float32.

https://github.com/ACEsuit/mace/blob/88d49f9ed6925dec07d1777043a36e1fe4872ff3/mace/calculators/mace.py#L145

error messages are not helpful so will likely take users time to troubleshoot this issue when encountered. only current workaround is to manually reset default dtype to float32 with 

```py
torch.set_default_dtype(torch.float32)
```

after every time MACE is called.

### Suggested fix

only convert model inputs to model's dtype without modifying _all_ float tensors everywhere

### minimal example

```py
import torch
from ase.build import bulk
from mace.calculators import mace_mp

orig_dtype = torch.get_default_dtype()
print(f"{orig_dtype=}")
>>> orig_dtype=torch.float32

atoms = bulk("Cu") * (2, 2, 2)
atoms.calc = mace_mp(default_dtype="float64")
atoms.get_potential_energy()

new_dtype = torch.get_default_dtype()
print(f"{new_dtype=}")
>>> orig_dtype=torch.float64
```

hatemhelal:
FWIW you can use a context manager to run different models with different torch default dtype. There is an implementation in PR #310, commit: https://github.com/ACEsuit/mace/commit/80211fd1fe29f4aa6ce872805c32b2807b343930

PR #1009, opened 2025-06-09
Isolate dtypes

CompRhys:
This is a fairly hefty set of changes that remove the use of global dtype setting from MACE codebase. This is done by adding custom `to` methods to change the dtypes of buffers. `dtype` is also made into a kwarg for all dataset objects.

All tests pass apart from those that have expected values that I believe are seed dependent and something must have been affected here that changes those values.

A bunch of minor code consistency changes are also included in this PR. Notable things are using `tmp_path` fixtures rather than manually creating and cleaning tmp folders, using the cli commands for the cli tests, adding some directories to the tests to make it more clear which tests are cli tests and which are core, adding concurrency to tests to reduce ci minute waste on PRs pushed in quick succession, using consistent docstring styles, using `npt.assert_allclose` consistently

The one major remaining source of confusion is in the `test_calculators.py` where using float32 causes everything to return very different/wrong values.

Related: https://github.com/ACEsuit/mace/issues/877, https://github.com/ACEsuit/mace/issues/328


mace/modules/radial.py
RokasEl, line 1: Same here, with the `to` overrides. Here's an example of current behaviour:

```
from mace.modules.blocks import AtomicEnergiesBlock
import numpy as np
import torch

block = AtomicEnergiesBlock(atomic_energies=np.array([1.0, 2.0, 3.0]))
print(block.atomic_energies.dtype)
block = block.to(torch.float64)
print(block.atomic_energies.dtype)

>>>torch.float32
>>>torch.float64
```

mace/modules/utils.py
```diff
@@ -302,8 +302,12 @@ def extract_invariant(x: torch.Tensor, num_layers: int, num_features: int, l_max
 def compute_mean_std_atomic_inter_energy(
     data_loader: torch.utils.data.DataLoader,
     atomic_energies: np.ndarray,
+    dtype: Optional[torch.dtype] = None,
 ) -> Tuple[float, float]:
-    atomic_energies_fn = AtomicEnergiesBlock(atomic_energies=atomic_energies)
+    dtype = dtype or torch.get_default_dtype()
+    atomic_energies_fn = AtomicEnergiesBlock(atomic_energies=atomic_energies).to(
+        dtype=dtype
+    )
```
RokasEl, line 310: I like this way of setting the dtypes, but I don't think we need to override the `to` methods everywhere.

mace/modules/wrapper_ops.py
```diff
@@ -150,8 +150,6 @@ def __new__(
                 layout=cueq_config.layout,
                 shared_weights=shared_weights,
                 internal_weights=internal_weights,
-                dtype=torch.get_default_dtype(),
-                math_dtype=torch.get_default_dtype(),
             )
         if (
             OEQ_AVAILABLE
```
RokasEl, line 155: I'm pretty sure the dtype is necessary here, as the kernels might be different depending on dtype

mace/tools/model_script_utils.py
```diff
@@ -67,7 +71,7 @@ def configure_model(
 
     elif (args.mean is None or args.std is None) and args.model != "AtomicDipolesMACE":
         args.mean, args.std = modules.scaling_classes[args.scaling](
-            train_loader, atomic_energies
+            train_loader, atomic_energies, dtype=dtype
         )
```
RokasEl, line 75: Do these classes really need a dtype kwarg? Can't this be done with a `cls.to(dtype)`?

mace/tools/torch_tools.py
```diff
@@ -72,7 +77,12 @@ def init_device(device_str: str) -> torch.device:
     return torch.device("cpu")
 
 
-dtype_dict = {"float32": torch.float32, "float64": torch.float64}
+dtype_dict = {
+    "float32": torch.float32,
+    "float64": torch.float64,
+    "": torch.get_default_dtype(),
+    None: torch.get_default_dtype(),
+}
```
RokasEl, line 85: I like this!

mace/data/lmdb_dataset.py
```diff
@@ -1,6 +1,10 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
CompRhys: MACE chooses to lint for py3.8 and non compatible type hints introduced later were used in this file

mace/cli/run_train.py
```diff
@@ -157,29 +163,36 @@ def run(args) -> None:
             model_foundation = torch.load(
                 args.foundation_model, map_location=args.device
             )
+            model_foundation = model_foundation.to(dtype=dtype)
```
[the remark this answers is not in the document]
CompRhys: this is again likely path dependent from trying to get it to work in that I was littering these in lots of places, will try find the right place for a single call.

mace/modules/wrapper_ops.py
CompRhys: this does turn out to be problematic which is maybe poor design from cuequivariance (especially given https://github.com/PASSIONLab/OpenEquivariance/issues/118 was responsive and implemented a working .to method that does update the kernels according to maintainers)

CompRhys:
This is blocked by getting a solution to https://github.com/NVIDIA/cuEquivariance/issues/124 and then updating the values in various tests that assert against seeded answers.

CompRhys:
`.to` method has been added in 0.6.0 cueq release so will try revive this at the weekend. It will require updating the pin to be `>0.6.0` for cueq which may cascade other issues.

PR #1048, opened 2025-07-02
dynamically set backend for distributed training based on device type

### LINKED ISSUE #1046: Multigpu training with intel GPUs (xpu) does not work
Dear @ilyes319, I have been trying to use `Intel Data Center GPU Max 1550` gpus for training /fine-tuning mace models. In the current implementation of code, one can run it on single intel GPU, but when trying to run with Multiple GPUs training gets stuck. Below are steps needed to reproduce this issue and also I propose possible solutions I think can fix it.

I have currently installed mace from the git repository directly (so all the recent changes are already included, but still the issue persists)

**Describe the bug**
When using Intel GPUs for training MACE with distributed settings enabled, training gets stuck.

**To Reproduce**
run `mace_run_train  with `--distributed` and `--device=xpu`

**Identified possible fixes/reasons**

1.  Also need to add this line  `import oneccl_bindings_for_pytorch`  here as per [docs](https://intel.github.io/intel-extension-for-pytorch/xpu/2.1.30+xpu/tutorials/features/DDP.html)https://github.com/ACEsuit/mace/blob/d39cc6b5f0f416dbc5eb3462f67544592130076e/mace/cli/run_train.py#L90  (needed to identify "ccl" as a valid backed for DDP, as "nccl" backend does not work with intel gpus)
2. This also needs to be dynamically set (if using xpu as device , it should be `torch.xpu.set_device(local_rank)`)https://github.com/ACEsuit/mace/blob/d39cc6b5f0f416dbc5eb3462f67544592130076e/mace/cli/run_train.py#L105
3.  Setting of Backend should be also be dynamically set here in this line https://github.com/ACEsuit/mace/blob/d39cc6b5f0f416dbc5eb3462f67544592130076e/mace/tools/distributed_tools.py#L50 

I did try these three changes as mentioned above locally and then training seems to have started successfully. It is very likely I may have overlooked something.  Would be great if anyone else can have a look and provide feedback on this if I have overlooked anything. I would be happy to raise a PR with the fixes.

naik-aakash:
Closes #1046 

## Changes

- Implemented 3 fixes proposed in the above issue.

### DIFF
--- a/.github/workflows/unittest.yaml
+++ b/.github/workflows/unittest.yaml
@@ -12,7 +12,7 @@ jobs:
         python-version: ['3.8','3.9','3.10','3.11','3.12','3.13']
 
     steps:
-      - name: checkout 
+      - name: checkout
         uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
@@ -31,7 +31,7 @@ jobs:
 
       - name: Run general unit tests
         run: |
-          pytest tests --ignore=tests/test_cueq.py
+          pytest tests --ignore=tests/test_cueq_oeq.py
 
   pytest-cueq:
     runs-on: ubuntu-latest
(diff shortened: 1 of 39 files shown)


ilyes319:
Looks good thank you!

PR #1071, opened 2025-07-14
Add option to plot interaction energy

CheukHinHoJerry:
In many cases plotting the energy without E0s instead of the total energy in the scatter plot is helpful. This PR creates an option to do so via an additional argparser

```python
    parser.add_argument(
        "--plot_interaction_e",
        help="Whether to plot energy without E0s",
        type=str2bool,
        default=False,
    )
```

And over the same data set, if `plot_interaction_e=False`, 
image

if `plot_interaction_e=True`, 
image

Not entirely sure this is the best way to add this functionality, please let me know if I should make any further changes.

### DIFF
--- a/mace/cli/run_train.py
+++ b/mace/cli/run_train.py
@@ -749,7 +749,8 @@ def run(args) -> None:
                 device=device,
                 plot_frequency=args.plot_frequency,
                 distributed=args.distributed,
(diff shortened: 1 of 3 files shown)


gabor1:
It would be great if both version would be plotted by default. they help diagnose different problems.

PR #1079, opened 2025-07-18
Updated MACE mu alpha

Nilsgoe:
I updated the old `MACE/mu_alpha` branch to the current main branch. It passes all tests of the current main branch.
In addition to the `AtomicDipolesMACE` and `DipoleMACE` models, it now also includes the `AtomicDielectricMACE` and `DipolePolarizabilityMACE` models. This allows users to train and calculate polarizabilities and dipoles. I am aware that there is some redundancy in having both sets of models in the code, but for backward compatibility, it might be useful to keep the old models available. Also one can then still have a L=1 model for the dipoles and doesn't need to train a L=2 model.

I tested the model with different SPICE subsets, and the training and prediction works fine.
I can also add a brief script on how to obtain the polarizabilities using ASE via `.get_properties["polarizability"]`, as well as some documentation on how to properly train these models.

It is still missing some new features, like the output plots, which one might want to add in later stages.

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -386,7 +386,11 @@ class MACECalculator(Calculator):
                 ret_tensors["forces"][i] = out["forces"].detach()
                 if out["stress"] is not None:
                     ret_tensors["stress"][i] = out["stress"].detach()
-            if self.model_type in ["DipoleMACE", "EnergyDipoleMACE", "DipolePolarizabilityMACE"]:
+            if self.model_type in [
+                "DipoleMACE",
+                "EnergyDipoleMACE",
+                "DipolePolarizabilityMACE",
+            ]:
                 ret_tensors["dipole"][i] = out["dipole"].detach()
             if self.model_type == "DipolePolarizabilityMACE":
                 ret_tensors["charges"][i] = out["charges"].detach()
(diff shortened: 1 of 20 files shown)


ilyes319:
@Nilsgoe Amazing, i ll have a look and merge. Thank you very much.

I think it would be nice to add some of this PR specific tests before merging. 
One test of a full training of the AtomicDielectricMACE, some calculators tests, some parsing of the new kind of data too.

nickhine:
Just saw that the mu_alpha branch was merged, which is good to see, as we have made quite a lot of use of branches derived from this code, so hopefully it will become easier to maintain those now.

However, a couple of things to note @Nilsgoe @ilyes319 :

There is, in this code as it stands, a significant performance hit manifesting as a slowdown as training of polarizability models proceeds. recently I tracked this down to the computational graph increasing with each epoch as a result of the creation of a new CartesianTensor() and its attachment to pytorch's computational graph each time the polarizability was transformed from spherical to cartesian. In the current version of my fork I have got rid of this by caching the CartesianTensor and the associated reduced_tensor_products within the training loop so that only one is created. Hopefully this might be of interest.

Secondly, Anas and I have successfully used extended this code to train models of polarizability and dipole using their derivatives, as calculated via DFTP (Z* and raman tensors) for the purpose of Raman Spectroscopy. We are writing
this up now and have been able to show that the use of derivatives significantly improves the accuracy of the models. I'd be keen to get this merged at some point (will get to work rebasing our branch on the current mace main branch to make that feasible.

Nilsgoe:
@nickhine First of all, it is great that you are using the code and I am looking forward to see the improvements when using derivatives. Secondly, one does not have to worry about the performance loss anymore, I fixed that before I did the pull requests. There were a couple of minor errors in the branch which I fixed and therefore the merged code does not have these anymore.

ilyes319:
@nickhine Thank you very much for your feedback. We would love to have a PR with the new functionalities and very much looking forward to your new paper on that.

PR #934, opened 2025-04-26
Added the option to include long-range interactions using Latent Ewald Summation(LES) method

BingqingCheng:
Added MACELES model. The short-ranged part is a standard MACE model, and we added long-range part using the LES library (https://github.com/ChengUCB/les).

To use:
- training: just specify --model="MACELES" when doing `python mace/scripts/run_train.py`
- evaluate: in addition to energy/forces/stress, one can predict Born effective charges on each atom. e.g.
python ~/repository/mace/scripts/eval_configs.py --configs ../fit-dipep-mace-les/spice-dipep-dipolar_val.xyz --output test-mace.xyz --model SPICE_small.model --batch_size 8 --default_dtype float32 --compute_bec

PR #1086, opened 2025-07-25
Reimplementation of MACELES

RokasEl:
This PR re-implements the MACELES model , updating it for compatibility with the current develop branch. Original PR #934 by @BingqingCheng.

## Verification and Changes
The new implementation has been checked against the old one. It initializes identically, and forward passes on the same data with freshly initialized models produce the same results.

A key correction is the inclusion of the `les_readouts` parameters in the optimizer. These parameters were previously not part of the parameter groups and were therefore not being trained. This PR corrects that. The new implementation also supports initializing a MACELES model from a foundation model checkpoint.

## Questions
- Given that `les_readouts` were not trained, was it the original intention for the latent charges to be derived from a fixed random projection of the node features? With this change, the projection will now be trained. @BingqingCheng
- You can see that in the test I've added, the model predicts a non-zero energy for isolated atoms even though `remove_self_interactions` is set to True in the Ewald block of LES (which is the default). Is this the expected behaviour? @BingqingCheng
- I'm not sure if want to keep the ability to finetune the foundation models, I would guess the optimal parameters will be quite different for a model with long-range effects and for a model without. @ilyes319, I am happy to remove the changes that enable this behaviour.

### DIFF
--- a/mace/cli/eval_configs.py
+++ b/mace/cli/eval_configs.py
@@ -54,7 +54,7 @@ def parse_args() -> argparse.Namespace:
     )
     parser.add_argument(
         "--compute_bec",
-        help="compute BEC",,
+        help="compute BEC",
         action="store_true",
         default=False,
     )
@@ -121,10 +121,8 @@ def run(args: argparse.Namespace) -> None:
 
     # Load model
     model = torch.load(f=args.model, map_location=args.device)
-    if model.__class__.__name__ != 'MACELES' and args.compute_bec:
-        raise ValueError(
-            "BEC can only be computed with MACELES model. "
-        )
+    if model.__class__.__name__ != "MACELES" and args.compute_bec:
+        raise ValueError("BEC can only be computed with MACELES model. ")
     if args.enable_cueq:
         print("Converting models to CuEq for acceleration")
         model = run_e3nn_to_cueq(model, device=device)
@@ -173,7 +171,11 @@ def run(args: argparse.Namespace) -> None:
 
     for batch in data_loader:
         batch = batch.to(device)
-        output = model(batch.to_dict(), compute_stress=args.compute_stress, compute_bec=args.compute_bec)
+        output = model(
+            batch.to_dict(),
+            compute_stress=args.compute_stress,
+            compute_bec=args.compute_bec,
+        )
         energies_list.append(torch_tools.to_numpy(output["energy"]))
         if args.compute_stress:
             stresses_list.append(torch_tools.to_numpy(output["stress"]))
@@ -257,12 +259,8 @@ def run(args: argparse.Namespace) -> None:
         assert len(atoms_list) == stresses.shape[0]
 
     if args.compute_bec:
-        bec_list = [
-            becs for sublist in bec_list for becs in sublist
-        ]
-        qs_list = [
-            qs for sublist in qs_list for qs in sublist
-        ]
+        bec_list = [becs for sublist in bec_list for becs in sublist]
+        qs_list = [qs for sublist in qs_list for qs in sublist]
 
     if args.return_contributions:
         contributions = np.concatenate(contributions_list, axis=0)
--- a/mace/cli/eval_configs.py
+++ b/mace/cli/eval_configs.py
@@ -53,6 +53,12 @@ def parse_args() -> argparse.Namespace:
         default=False,
     )
     parser.add_argument(
+        "--compute_bec",
+        help="compute BEC",,
+        action="store_true",
+        default=False,
+    )
+    parser.add_argument(
         "--return_contributions",
         help="model outputs energy contributions for each body order, only supported for MACE, not ScaleShiftMACE",
         action="store_true",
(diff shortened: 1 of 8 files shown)


RokasEl:
~~Todo for me, remove the `**kwarg` from mace/scaleshiftmace since this breaks lammps integrations. That was just a convenience to reduce the number of changes for the eval_configs script.~~

Edit: now done.

BingqingCheng:
That's fantastic. I'll try out the new implementation in the next days!


   - Given that les_readouts were not trained, was it the original
   intention for the latent charges to be derived from a fixed random
   projection of the node features? With this change, the projection will now
   be trained. @BingqingCheng <https://github.com/BingqingCheng>

The intention is to train the les_readouts. I'm surprised to hear that they
were not traind, as all our MACELES models trained on different systems
have correct charges. I'll double check on this. But anyway, if it's true,
now we know that even a random project works fine...

   - You can see that in the test I've added, the model predicts a non-zero
   energy for isolated atoms even though remove_self_interactions is set to
   True in the Ewald block of LES (which is the default). Is this the expected
   behaviour? @BingqingCheng <https://github.com/BingqingCheng>

Do you mean the energy of a system with just one isolated atom in a
periodic box? In that case the charge is not neutral and a uniform
neutralizing background charge is implicitly added during the Ewald
summation. The "remove_self_interaction" removes the self-interaction of
this charge, but not the energy due to the background. So yes, there will
be non-zero energy for isolated atoms. In principle, it's easy to also just
add a line to remove the background energy, although I don't think there's
much practical use there. I can add it if you think it's needed.

   - I'm not sure if want to keep the ability to finetune the foundation
   models, I would guess the optimal parameters will be quite different for a
   model with long-range effects and for a model without. @ilyes319
   <https://github.com/ilyes319>, I am happy to remove the changes that
   enable this behaviour.

RokasEl:
>The intention is to train the les_readouts. I'm surprised to hear that they
were not traind, as all our MACELES models trained on different systems
have correct charges. I'll double check on this. But anyway, if it's true,
now we know that even a random project works fine...

Yes, this is easy to miss. The MACE `run_train` script creates multiple specialised parameter groups rather than just passing the full `model.parameters()`. If you used your own training script rather than the MACE one, then you might have been training the readouts.

> Do you mean the energy of a system with just one isolated atom in a
periodic box? In that case the charge is not neutral and a uniform
neutralizing background charge is implicitly added during the Ewald
summation. The "remove_self_interaction" removes the self-interaction of
this charge, but not the energy due to the background. So yes, there will
be non-zero energy for isolated atoms. In principle, it's easy to also just
add a line to remove the background energy, although I don't think there's
much practical use there. I can add it if you think it's needed.

Thanks for the explanation! I don't think we need to add this feature, I just waned to double check.

pb1414:
Thank you very much for the LES implementation and updates! Really appreciate this! @RokasEl , is there by chance a reason the new implementation uses notably more memory? Comparing with pr-934, les_readout parameters are now smoothly learned using run_train, but it's using about 1-2 orders of magnitude more memory than MACE/ScaleShiftMACE/old LES on my GPU, causing OOMs with moderate batch sizes. Wrapping les in torch.no_grad() and removing its energy contribution cuts memory use, but only by about half. I will investigate this further. Thank you again!

Edit: Found issue. The _periodic_ Ewald summation method is activated if the input structure is defined in a unit cell, even if all three PBC parameters are false. This blew up memory consumption because my cell was arbitrarily large. Could be an issue with my training data file, but may submit a tiny pull request to pass a periodicity parameter.

RokasEl:
@pb1414, it should definitely increase the memory usage as there are now both more parameters in the model and you have to store more gradients (now that the readouts are actually being trained). However, I am very surprised this would increase the memory by  orders of magnitude rather than 10/20 per cent!

kevinkong98:
Any chance this would support cueq acceleration soon?

RokasEl:
@kevinkong98, it should support cueq as it is. If you're getting errors let me know and I'll take a look.

kevinkong98:
I got
2025-07-31 18:57:33.914 INFO: Converting model to CUEQ for accelerated training
  File "/home/fslcollab457/miniconda3/envs/MACE_LES/bin/mace_run_train", line 8, in <module>
    sys.exit(main())
  File "/home/fslcollab457/miniconda3/envs/MACE_LES/lib/python3.10/site-packages/mace/cli/run_train.py", line 75, in main
    run(args)
  File "/home/fslcollab457/miniconda3/envs/MACE_LES/lib/python3.10/site-packages/mace/cli/run_train.py", line 687, in run
    assert model.__class__.__name__ in ["MACE", "ScaleShiftMACE"]
AssertionError


And if I change the relevant code in mace_run_train to include MACELES

I get

  File "/home/fslcollab457/miniconda3/envs/MACE_LES/bin/mace_run_train", line 8, in <module>
    sys.exit(main())
  File "/home/fslcollab457/miniconda3/envs/MACE_LES/lib/python3.10/site-packages/mace/cli/run_train.py", line 75, in main
    run(args)
  File "/home/fslcollab457/miniconda3/envs/MACE_LES/lib/python3.10/site-packages/mace/cli/run_train.py", line 688, in run
    model = run_e3nn_to_cueq(deepcopy(model), device=device)
  File "/home/fslcollab457/miniconda3/envs/MACE_LES/lib/python3.10/site-packages/mace/cli/convert_e3nn_cueq.py", line 192, in run
    num_product_irreps = len(config["hidden_irreps"].slices()) - 1
KeyError: 'hidden_irreps

Many thanks!!

RokasEl:
@kevinkong98, PR #1113 will fix this once merged.

ilyes319:
@RokasEl merged

PR #1095, opened 2025-07-28
use the one available head if that's all there is, regardless of name

bernstei:
automatically use the head if one is all that's all that's available, even if it's not named "default"


bernstei:
@ilyes319 I think you might as well merge this one.  The small model problem is unrelated, and we'll deal with it separately.

PR #1098, opened 2025-07-28
don't apply user's --atomic_numbers to pt_head with doing multihead

### LINKED ISSUE #1097: develop branch multihead replay doesn't end up with correct set of atomic numbers for pt model
configuration selection for multihead replay in `run_train.py` requires the user pass `--atomic_numbers` to determine the atomic numbers that will be used to select configurations (via strategies combinations, inclusive, etc).  However, run_train constructs `head_config` for all heads by calling `dict_head_to_dataclass`, which initializes `head_config.atomic_numbers` to `args.atomic_numbers` in https://github.com/ACEsuit/mace/blob/7ad83d0569c9a31e0c288c48340afde0a5d1f4be/mace/tools/multihead_tools.py#L73 and it does this for all heads

As a result, all heads (including the pt) already have their atomic number list to that requested by the user for the real model.  I suspect that this shouldn't be set for the pt_head, or be reset back to `None`, so that when it gets to https://github.com/ACEsuit/mace/blob/7ad83d0569c9a31e0c288c48340afde0a5d1f4be/mace/cli/run_train.py#L368 it will set them from the actual configurations.

bernstei:
Currenly `run_train.py` applies the user's `--atomic_numbers` to select the atomic numbers that will be supposed in all heads.  This wipes that setting for the "pt_head" if multihead replay is active, so that the replay head gets its atomic numbers from the set included in its fitting data, which could include other elements, e.g. if random padding is used to add more configurations beyond the element-specific filtering to reach the request number `--num_samples_pt`. 

closes #1097

### DIFF
--- a/mace/cli/run_train.py
+++ b/mace/cli/run_train.py
@@ -226,6 +226,10 @@ def run(args) -> None:
     for head, head_args in args.heads.items():
         logging.info(f"=============    Processing head {head}     ===========")
         head_config = dict_head_to_dataclass(head_args, head, args)
+        # don't apply user's --atomic_numbers to pt_head, that info needs to come
+        # from the actual pt data
+        if args.multiheads_finetuning and head_config.head_name == "pt_head":
+            head_config.atomic_numbers = None
 
         # Handle train_file and valid_file - normalize to lists
         if hasattr(head_config, "train_file") and head_config.train_file is not None:


bernstei:
Working on a test for this, but running into software engineering issues.  Namely, there code that filters the pretrained data is currently only used if the "special" `--pt_train_file=mp` is used - it's entirely distinct, and not easily compatible, with the code that reads a user-specified actual pt training file, because the latter is just the generic "read data for a head" code.

For the purpose of this test I can create a dummy `~/.cache/mace/mptraj.xyxz` (or whatever it's called), but I can't decide if that's sufficient, or if we should refactor the reading code so that the filtering can be applied to any pt_head training file.

bernstei:
I did the test with a dummy cached "mp" data file for now.  I'll think about refactoring so the filtering can be applied to arbitrary replay data files, not just the special "mp" auto-downloaded one.

bernstei:
@ilyes319 I think we should merge this whenever you're satified and leave any refactoring of the filtering code for another day.

bernstei:
If there's something I can do to help merge this into develop before develop is merged into main, please let me know.

ilyes319:
looks good to me, I will merge that.

PR #1102, opened 2025-07-30
mace.tools.train.MACELoss Metric filter out quantities whose weight was 0 during training

### LINKED ISSUE #939: mace_run_train reports crazy error if fitting set contains configs that have no reference data
It appears that configs that have no reference data (not sure if this is also true if they have, e.g., forces but no energy), seem to use "0" for that value when calculating the errors printed on the `Epoch: ...` lines.  Example is below. It's not so obvious on the Epoch lines for this particular toy example, because the fit is so terrible (because it's fitting to ASE EMT data), but it's obvious in the `results/*png` parity plot.

```
mace_run_train --multiheads_finetuning false --foundation_model small --energy_weight 10 --forces_weight 10 --stress_weight 1 --ema  --ema_decay 0.9999 --force_mh_ft_lr true --lr 0.0001 --max_num_epochs 1 --train_file fit_2.xyz --valid_fraction 0.2 --default_dtype float64 --name MACE.test --device cpu
```
(rename fitting data file to `.xyz`)

[fit_2.txt](https://github.com/user-attachments/files/19967421/fit_2.txt)

Image

bernstei:
Remove quantities from error table and parity plot if they have 0 weight.

Based on weight code from training-loss functionality in https://github.com/ACEsuit/mace/blob/d39cc6b5f0f416dbc5eb3462f67544592130076e/mace/modules/loss.py#L132 but refactored to apply more cleanly to metrics data structures.

[added] Refactored from initial version to also cleanly apply to parity plot

[probably still needs tests]

closes #939

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -138,6 +138,13 @@ class MACECalculator(Calculator):
                 self.compute_atomic_stresses = True
         elif model_type == "DipoleMACE":
             self.implemented_properties = ["dipole"]
+        elif model_type == "DipolePolarizabilityMACE":
+            self.implemented_properties = [
+                "charges",
+                "dipole",
+                "polarizability",
+                "polarizability_sh",
+            ]
         elif model_type == "EnergyDipoleMACE":
             self.implemented_properties = [
                 "energy",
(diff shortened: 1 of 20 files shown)


bernstei:
I'm working on the dipole test failures.

bernstei:
I'll look at the conflict in the next couple of hours

PR #1109, opened 2025-07-31
Fix test of real to pretrain data ratio for repeating configs

### LINKED ISSUE #1108: test of size of real heads to pt_head is broken
The test that's supposed to decide if the real head data should be repeated to bring it closer to the size of the pt_head replay data
https://github.com/ACEsuit/mace/blob/1c2eb1554ef7d3ff8cc9d1579a440bf2c0cada6b/mace/cli/run_train.py#L354 is broken.  Instead of checking size(real heads) / size(pt head) it checks size(all heads) / size(pt head).

I'll open a PR to fix this once we decide how to fix it.  Compare sum(real heads) / size(pt_head)?  Or for _each head_ separately check if it's too small and repeat its data?

~~Also, should we set it so it always takes integer multiples of the real head sets, so some configs aren't weighted more than others within that set?~~ Never mind - I see the `int()` call.

bernstei:
Correct numerator in ratio of real to pretrain configs when checking ratio to see if duplicating real configs is needed.

Also (incidentally) remove unneeded debugging prints.

closes #1108

### DIFF
--- a/mace/cli/run_train.py
+++ b/mace/cli/run_train.py
@@ -352,16 +352,16 @@ def run(args) -> None:
         assert head_config_pt is not None, "Pretraining head not found"
         if all_ase_readable:
             ratio_pt_ft = (size_collections_train - len(head_config_pt.collections.train)) / len(head_config_pt.collections.train)
-            if ratio_pt_ft < 0.1:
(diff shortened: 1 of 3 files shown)


bernstei:
Working on a command line flag to set the threshold, so 0 or negative to disable the data repeat.

PR #1121, opened 2025-08-07
Hotfix: Solving weight shape error for polarizability_weights

Nilsgoe:
There was a small bug in the training of the polarizability, where the dimensions of the polarizability_weight had not been properly added with regard to some other updates made in the version update. Unfortunately, this bug prevented the model from training at all.

@ilyes319 I would highly recommend merging this directly into the main branch; otherwise, the training on polarizabilities won't be supported at all

### DIFF
--- a/mace/data/atomic_data.py
+++ b/mace/data/atomic_data.py
@@ -239,9 +239,17 @@ class AtomicData(torch_geometric.data.Data):
             )
             if config.property_weights.get("polarizability") is not None
             else torch.tensor(
-                1.0, dtype=torch.get_default_dtype()
-            )  ### Might need to be updated
+                [[1.0, 1.0, 1.0], [1.0, 1.0, 1.0], [1.0, 1.0, 1.0]],
+                dtype=torch.get_default_dtype(),
+            )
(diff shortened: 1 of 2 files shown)


ilyes319:
Will merge that thank you!

ilyes319:
Curious was that did not make the test to fail? can we update the test too.

Nilsgoe:
Currently, neither  `AtomicDipolesMACE` nor  `AtomicDielectricMACE` is part of `test_run_train.py`. Both are only tested in `test_calculator.py`, which I suspect is the problem. When adjusting `AtomicDielectricMACE`, `AtomicDipolesMACE` served as a model, but in the `trained_dipole_fixture` function no `dipole_weight` is set. Similarly, in the `trained_dipole_polar_fixture` function neither `dipole_weight` nor `polarizability_weight` is set. I haven’t tested it, but I guess adding those weights in `test_calculator.py` and including tests for both models in `test_run_train.py` would have spotted the error.

ilyes319:
yes we would need a run train test for the model.

PR #1129, opened 2025-08-11
Adding training test to AtomicDipolesMACE and AtomicDielectricMACE.

Nilsgoe:
I added for each model (`AtomicDipolesMACE` and `AtomicDielectricMACE`) a training test in `test_run_train.py`.

If more testes are needed, we can add those to this PR.


Nilsgoe:
I had to make a separate `test_run_train_dipole_polar.py` file. Somehow, the  `test_run_train.py` was failing for some other tests when I added additional properties to the atoms.info (like  `c.info["REF_polarizability"] = np.random.normal(0.1, size=(3,3))`)

PR #1134, opened 2025-08-14
MACELES: Ensuring correct LES energy method is called based on boundary conditions

pb1414:
**Problem:**
The current MACELES method seems to call the periodic Ewald summation to calculate les_energy even when the system is finite, rather than the [intended pairwise direct sum](https://arxiv.org/pdf/2507.14302). For example, this happens when training on the [MACE-OFF23 xyz files](https://www.repository.cam.ac.uk/items/d50227cd-194f-4ba4-aeb7-2643a69f025f).

**Proposed Solution:**
The LES library decides which method to use based on "if cell is None or torch.linalg.det(cell[i]) < 1e-6:". Many solutions could work well, but this PR sets the corresponding LES cell input argument entries to zero (thus zero determinant) when all three PBC parameters for the configuration are False. 

**Testing:**
-Results (and memory usage) match expected values corresponding to correct les_energy method
-Appears to have negligible impact on the duration of the MACELES forward pass (briefly tested this)
-Does not affect performance of other methods (edits are primarily in MACELES forward pass)

**Alternatives:**
I attempted to make the change during data preprocessing, but fixes took too long for large datasets. I'm happy to explore other options if this one isn't great. 

Thank you for maintaining MACE!


ilyes319:
@pb1414 Thanks for that! You should make the pbc a torch tensor of bools and not a list.

pb1414:
Thank you, just made the edits and rechecked performance!

ilyes319:
@pb1414 Great thank you!

PR #1158, opened 2025-09-01
Docs: add polarizability example and Hessian citation

Nilsgoe:
I added some documentation on how one could train an AtomicDielectricMACE and how to get the polarizability with ASE. I also added the citation for the Hessian paper to the respective page in the documentation.


ilyes319:
great thank you

PR #1162, opened 2025-09-03
Fix: Old AtomicDielectricMACE models trained only on Dipoles were not properly supported.

Nilsgoe:
Now, `AtomicDielectricMACE` models from legacy versions of the code are also supported. For example, models from [https://github.com/venkatkapil24/MACE-OFF-mu-models](https://github.com/venkatkapil24/MACE-OFF-mu-models). 

Credits to @cecicxy, who reported the problem in this [issue](https://github.com/venkatkapil24/MACE-OFF-mu-models/issues/2) .


ilyes319:
Nice thanks!

PR #1179, opened 2025-09-12
[DRAFT] Fix MACECalculator.results dict property names

### LINKED ISSUE #1178: MACECalculator sets self.results field incorrectly
self.results["energies"] is supposed to be the site (atom) energy, i.e. more like `"node_energies"`, rather than, apparently, the energies from the ensemble https://github.com/ACEsuit/mace/blob/42cd495b9c8e155441c405fef842a8ba5107c2f8/mace/calculators/mace.py#L484

I think "energies" should be "energy_comm", in analogy with "forces_comm" which is used now, and "energies" needs to contain something related to "node_energies"

bernstei:
Make sure keys in `results` dict are consistent with each other and with ASE standards.

Generally refactor code that sets calculated quantities

closes #1178

### DIFF
--- a/mace/calculators/mace.py
+++ b/mace/calculators/mace.py
@@ -152,15 +152,16 @@ class MACECalculator(Calculator):
                 f"Give a valid model_type: [MACE, DipoleMACE, DipolePolarizabilityMACE, EnergyDipoleMACE], {model_type} not supported"
             )
 
+        # superclass constructor initializes self.implemented_properties to an empty list
         if model_type in ["MACE", "EnergyDipoleMACE"]:
-            self.implemented_properties = [
+            self.implemented_properties.extend([
                 "energy",
                 "energies",
                 "free_energy",
                 "node_energy",
                 "forces",
                 "stress",
-            ]
+            ])
             if kwargs.get("compute_atomic_stresses", False):
                 self.implemented_properties.extend(["stresses", "virials"])
                 self.compute_atomic_stresses = True


bernstei:
Inconsistencies that still remain as of 34c8aad:
- `virials` but not `virial` (`stress` and `stresses` both exist)
- random mix of `_comm` and `_var` for different properties when ensemble is used

bernstei:
Tests are passing now, including the new one I did for the atomic energies.   Is that enough, i.e. do we trust the tests, or do we want further review of my changes?

ilyes319:
looks good, can you solve the develop branch conflict, than i will merge.

bernstei:
> looks good, can you solve the develop branch conflict, than i will merge.

I think that's fixed now, but don't merge yet.  I have a question.

bernstei:
As I was fixing the conflict, I started wondering if all the lines like https://github.com/ACEsuit/mace/blob/3959ad0409332328207752fae41897e18beab376/mace/calculators/mace.py#L449-L455 (which copy from `out` to `ret_tensors`) should actually be refactored.  Instead of testing on the `model_type` and then, for some fields, testing if they are not None, what about setting every possible field in `ret_tensors` if it's not None in `out`, regardless of the model type?  I think that'll be more robust to later changes in models, no?

along the lines of 
```
if out.get("energy") is not None:
   ret_tensors["energy"] = out["energy"][i].detach()
```

bernstei:
I guess the one downside is that if a value is missing from `out` there won't be a `KeyError` at this stage, only later, when the user tries to get that property.  But I feel like that shouldn't really happen ever, and we could make unit tests that are model_type-specific to make sure that `out` contains what we expect.

bernstei:
More consistency questions - is there a reason `node_energy` isn't converted to eV?

bernstei:
Code can be refactored to being much shorter if we convert from `out` to `ret_tensors` and `ret_tensors` to `self.results` without explicitly referring to `model_type`.  I'm going to push a commit, and we can revert if you don't like it.

ilyes319:
I think it should be converted. mmm I see what you mean I think I like it.

bernstei:
@ilyes319 can you explain why atomic stresses in `ret_tensors` are created "on the fly" in https://github.com/ACEsuit/mace/blob/b91bbc2c88fa1ecddcea19f3faa5fce4b1019bbb/mace/calculators/mace.py#L465-L467
while every other quantity in `ret_tensors` is created as 0 torch tensors first, in `_create_result_tensors()` https://github.com/ACEsuit/mace/blob/b91bbc2c88fa1ecddcea19f3faa5fce4b1019bbb/mace/calculators/mace.py#L345-L355

Is creating the 0 tensors actually useful, or should this refactor just skip that step if that makes the code is cleaner?

bernstei:
I pushed a commit that implements this refactor, but really it could be a lot simpler if the `torch.zero(...)` calls could be skipped. I don't understand why they are necessary given that saving `atomic_stresses` in `ret_tensors` never used them.

[edited] I guess creating the zero tensors with a leading dimension of `num_models` and then copying each model's value into a separate slot makes it possible to call things like `torch.mean` on the entire assemblage, which is what happens now for most of the quantities, but doesn't work if you append them into a _list_ of torch tensors, which is what happens to `atomic_stresses`.  Given that, I guess the current version is probably correct, or we could convert to a list of cpu numpy objects, model by model, then use `np.mean` and `np.var` to do the ensemble analysis.

bernstei:
One thing we could still do to simplify is to make `_create_result_tensors` take the first model's `out` object as its argument, and construct the result tensor dict based on the quantities defined in that structure, rather than a pre-existing list based on the `model_type`.  @ilyes319  Is that worth doing?

bernstei:
Now I think it's about as clean as it can get without major changes.  @ilyes319 take a look

alinelena:
since you are at this... can you test you can dump the object in extxyz in ase?

bernstei:
> since you are at this... can you test you can dump the object in extxyz in ase?

I'll check, but given that it's using non-ASE-standard additional properties, I'm not at all sure we should expect it to work.

bernstei:
> since you are at this... can you test you can dump the object in extxyz in ase?

What specifically are you expecting?  I get no errors writing, but the non-standard properties are not stored.

bernstei:
Tests are passing.  What do you think, @ilyes319, @alinelena ?

alinelena:
ok looks to work, we just need to document node_enegy does not have the e0s.

ilyes319:
@bernstei One thing is that now the Dipole model will have the "energy" in the implemented properties, while it does not return any. How safe is that?

bernstei:
> @bernstei One thing is that now the Dipole model will have the "energy" in the implemented properties, while it does not return any. How safe is that?

I didn't think I changed much of the logic for `implemented_properties`, but it's true that I wasn't focused on that.  Let me look.

bernstei:
> @bernstei One thing is that now the Dipole model will have the "energy" in the implemented properties, while it does not return any. How safe is that?

I looked, and don't understand why you think this.  The only place `"energy"` is added to `implemented_properties` is https://github.com/ACEsuit/mace/blob/07ff8c2e006a0a7328a2dbc1e2df833e630e643d/mace/calculators/mace.py#L155-L163 which won't happen if it's `DipoleMACE`.  

<del>I did, however, notice a bug where `DipoleMACE` will fail at https://github.com/ACEsuit/mace/blob/07ff8c2e006a0a7328a2dbc1e2df833e630e643d/mace/calculators/mace.py#L167-L168 <del>since it'll try to extend an array that wasn't declared, which I'll fix now, but I'm surprised that no test caught it.</del>

<del>Is there test coverage for every model type?</del>

Looks like self.implemented_properties starts out as an empty array due to the superclass constructor, so that's why it works.

ilyes319:
ok i thought it would start initialised with some default things. If it is empty that it is fine.

bernstei:
I'm comfortable merging, unless we think it needs another pair of eyes.

PR #1199, opened 2025-09-23
Revert setting requires grad in data pipeline

hatemhelal:
This should address the regression reported in #1196 

Note that this change requires that compiled mace users to explicitly set requires_grad on the positions input.  This is already handled within the ASE calculator wrapper.


alinelena:
@ilyes319 @hatemhelal I tested this and seems to fix the issue

ilyes319:
thank @hatemhelal @alinelena

PR #1222, opened 2025-10-09
[DRAFT] Embedding of arbitrary per atom quantities

bernstei:
Facilitate creation of models that depend on arbitrary per-atom quantities, categorical or continuous.

Categorical quantities are currently restricted to being dependent on `Atoms.arrays` quantities that are integers in the range `0:nclasses`.  One idea to extend this would require supporting syntax like
```
embedding_specs:
  categorical_desc:
    key: "array_key_cat:val0:val1:val2"
    type: categorical
    .
    .
```
In this PR you have to pass `MACECalculator(... arrays_keys={'categorical_desc': 'arrays_key_cat'}`), where `'categorical_desc'` needs to match the `embedding_specs` key that was used during fitting, and `arrays_key_cat` must be an `Atoms.arrays` key available at evaluate time, similarly to the way per-config embedding passes `info_keys` (https://github.com/ACEsuit/mace/blob/develop/tests/test_embedding_train.py#L151). With this extension, you would need to pass the same complex "key" (with the possible values embedded) to `MACECalculator` so the classes computed at fit time would match the classes computed at evaluate time.  An alternative would be to store, for each categorical embedding, the possible class values in the torch model.

Also, the approach used here might allow for significant simplification of the creation of the other (currently hard-coded) `AtomicData` tensors.

### DIFF
--- a/mace/data/utils.py
+++ b/mace/data/utils.py
@@ -71,7 +71,7 @@ def update_keyspec_from_kwargs(
             arrays_keys[key[:-4]] = keydict[key]
 
     # automagically add properties for embeddings
-    if "embedding_specs" in keydict:
+    if keydict.get("embedding_specs") is not None:
         for embed_name, embed_spec in keydict["embedding_specs"].items():
             key = embed_spec.get("key", embed_name)
             if embed_spec["per"] == "atom":


bernstei:
What do you think, @ilyes319 ? The actual guts of the change are pretty minimal, so I'm pretty happy with the design.  The only issue was the shape of the tensors stored in `AtomicData`, but it seems to be OK for embeddings.

The major refactor of the conversion from Atoms to Configuration to AtomicData would be pretty straightforward if the tensor shapes are predictable from whether they are per-config or per-atom, and whether they are scalars or higher dimension, but I haven't figured out if that predictability is in fact true (or at least mostly true, so we could do a general thing and special case a couple of specific things that violate the pattern).

ilyes319:
Hey @bernstei, looks fine to me, it seems quite minimal. I will merge for now but might slightly rework it before the next release.

PR #1234, opened 2025-10-20
note in docs that atom_modify map can be any valid lammps value, not just 'yes'

### LINKED ISSUE #1226: document atom_modify map for LAMMPSlib
It might be worth it to make it explicit in the mace mliap pair-style docs that it also works with `atom_modify map array`, which is the `LAMMPSlib` default, in addition to the `map yes` that's used in the examples.  The problem is that while overriding the `LAMMPSlib` (if you think you need to switch `LAMMPSlib` from array to yes) it's easy to forget the `sort 0 0` that's also required, and leads to weirdly wrong results.

bernstei:
Explicitly document `atom_modify map` values compatibility with `LAMMPSlib`

closes #1226

### DIFF
--- a/docs/guide/lammps_mliap.rst
+++ b/docs/guide/lammps_mliap.rst
@@ -130,6 +130,8 @@ The ``0`` after the model filename is a required parameter for the unified ML-IA
 
 The element list after ``pair_coeff * *`` should be ordered as you want them to appear in LAMMPS, and must be a subset of the elements your model was trained on.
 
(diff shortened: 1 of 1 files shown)


wcwitt:
Are you sure `atom_modify map` is actually required input for ML-IAP? I.e. you might be able to drop it entirely from these docs.

bernstei:
> Are you sure `atom_modify map` is actually required input for ML-IAP? I.e. you might be able to drop it entirely from these docs.

Good point - I guess it's easy enough to check

PR #1297, opened 2025-12-01
Add E0 re-estimation and LoRA improvements

ttompa:
## Add LoRA weight merging and improved E0 estimation for fine-tuning

This PR introduces two main improvements for model fine-tuning:

1. **LoRA weight merging**: After training with LoRA adapters, the learned weights are now automatically merged back into the base model weights before saving. This produces a standard model file without unnecessary overhead.

2. **Enhanced E0 estimation**: Added a new `--E0s="estimated"` option for fine-tuning foundation models. Instead of simple averaging, this estimates atomic reference energies by:
   - Running the foundation model on all training configurations
   - Computing prediction errors (reference - predicted energies)  
   - Solving a least-squares system to find optimal E0 corrections for each element
   
   This approach provides a better estimate for the E0s compared to the existing `"average"` method when isolated energies are not available.

Additional changes include LoRA delta caching in eval mode for faster training and additional unit tests.

### DIFF
--- a/mace/cli/run_train.py
+++ b/mace/cli/run_train.py
@@ -34,9 +34,9 @@ from mace.cli.convert_e3nn_oeq import run as run_e3nn_to_oeq
 from mace.cli.convert_oeq_e3nn import run as run_oeq_to_e3nn
 from mace.cli.visualise_train import TrainingPlotter
 from mace.data import KeySpecification, update_keyspec_from_kwargs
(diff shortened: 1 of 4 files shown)


ilyes319:
hey @ttompa thx for that! could you move the lora_tools.py to the modules folder. The philosophy is that heavy pytorch codes that changes things in the model belongs to the modules folder and the tools are more python or less model architecture code.

PR #1310, opened 2025-12-18
Write files only from root process when distributed

### LINKED ISSUE #1309: mace_run_train distributed processes overwrite each others' pretrained data files
There's no mechanism to ensure that pretrained data writing that happens as part of `mace_run_train` in distributed mode only occurs on one process.  As a result, it's possible for the different processes to overwrite each others' `mp_finetuning...xyz` files, and then fail when they are being read back in.

bernstei:
In particular, fixes issue where distributed mace_run_train processes overwrite each others' pretrained and combined data files. It's possible that there are other places where this kind of error exists, but I tried to fix it everyplace I could find it.

closes #1309

### DIFF
--- a/mace/cli/fine_tuning_select.py
+++ b/mace/cli/fine_tuning_select.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import argparse
 import ast
 import logging
+import time
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


bernstei:
includes #1311 , sort of accidentally

PR #1327, opened 2026-01-09
eval_configs: fix node energies bug, simplify force handling

ThomasWarford:
Fix node energies error (occurring when number of atoms varies between configs in batch),

Simplify force handling using `List.extend` rather than `List.append` + python list comprehension


ThomasWarford:
I added a few tests here, should be good to merge once #1374 is :) @ilyes319

Issue #1263, opened 2025-11-03
Can non-linear MACE models be used in LAMMPS (e.g. mace-mh-1)


foster362:
I'm wondering if there is currently a way to convert non-linear MACE models (e.g. mace-mh-1) into the lammps-iap format. I tried, mace_create_lammps_model mace-mh-1.model --format=mliap but get:

AttributeError: 'RadialMLP' object has no attribute 'hs'

pip list: mace-torch                    0.3.15

ilyes319:
Hey @foster362 I am working on it, will push soon.

ilyes319:
@foster362, can you try with the latest develop + redownloading the mace-mh-1 model (i just uploaded a modified one).

foster362:
Thanks for working on this. With the latest mace version (0.3.15) I'm able to create the non-linear lammps iap models but I'm getting an error trying to use the model in lammps:

AttributeError: 'MACEEdgeForcesWrapper' object has no attribute 'total_charge'
ERROR: Running mliappy unified compute_forces failure. (src/KOKKOS/mliap_unified_kokkos.cpp:82)

ilyes319:
mm did you make sure to run the lammps with the same mace as the compilation? could you send me the compiled model you get?

foster362:
Everything works if I use mh-0 but not when I use the mh-1 version.
Attached is the model. Thanks
 mace-mh-1.model-mliap_lammps.pt
<https://drive.google.com/file/d/1rd-K-Tej2RuB-LaBttlp2vt5qLSD9xMi/view?usp=drive_web>

foster362:
Any ideas what the issue is? If it works for you, I can try building mace/lammps with the same versions of compilers (e.g. python, cuda...)

ilyes319:
I have no idea what might be happening to be honest... I would need to dig that more. That error: AttributeError: 'MACEEdgeForcesWrapper' object has no attribute 'total_charge' is very confusing to me because I explicitly fixed it. I will share a debug branch with you if you are up to running it with it.

foster362:
I can give the debug branch a try.

skaiser29:
Hey, just wanted to see if the support for mace-mh-1 with mliap ever got sorted. I tried today with the latest commit on the dev branch. I'm able to compile the mace-mh-1 model without error, but at runtime, I get a bunch of pytorch errors from vectorized_gather_kernel ops, job log attached. 

[...]

PR #1331, opened 2026-01-16
Mliap nonlinear fix

skaiser29:
Addresses https://github.com/ACEsuit/mace/issues/1263. 

I tested that forces match the ASE implementation for mace-mh-1, dynamics are stable, and multi-gpu works correctly. 

Note: reliant on lammps develop newer than December 15th, 2025, due to https://github.com/lammps/lammps/pull/4808. Using older lammps releases will result in incorrect physics.


ilyes319:
hey @skaiser29 thank you for that!

Issue #1322, opened 2026-01-01
CI runs of of disc space


stenczelt:
Tests in the CI are sometimes running out of disc space, preventing the test run to complete. See https://github.com/ACEsuit/mace/actions/runs/20615258839/job/59206673218?pr=1321 for example for #1321 

<img width="1552" height="1463" alt="Image" src="https://github.com/user-attachments/assets/d46796ac-5733-4836-9b23-ff97209bb5dd" />

PR #1332, opened 2026-01-21
Fix1322 free up disk space in CI

stenczelt:
# summary

For #1322: the CI was running out of disc space. We are reclaiming disc space on the runner by deleting some large libraries irrelevant to the project

# details

We've fixed a symptom here, not the root cause: some tests are littering a lot of files and not cleaning up after themselves, ca 10GB total. 

We are freeing up 30GB space by deleting Java, .NET, Swift, Haskell, Android SDKs, and web browsers, which are not needed for this project's test but are default in the actions image.

Additionally, I've
- added logging of disc space & duration after each test
- fixed an int casting issue in the tests - this was failing post python 3.8, hidden before by the disk space issue I think
- tests to run on the `develop` branch as well

### DIFF
--- a/tests/test_calculator.py
+++ b/tests/test_calculator.py
@@ -10,7 +10,7 @@ import torch
 from ase import build
 from ase.atoms import Atoms
 from ase.calculators.test import gradient_test
-from ase.constraints import ExpCellFilter
+from ase.filters import FrechetCellFilter
 
 from mace.calculators import mace_mp, mace_off
 from mace.calculators.foundations_models import mace_omol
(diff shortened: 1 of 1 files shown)


ilyes319:
@stenczelt Thank you! I was wondering also if you could have a look at the pytest group jobs, so we can parallelism more the tests, and make them run faster. @jameskermode mentioned he is using it in matscipy.

jameskermode:
See https://github.com/libAtoms/matscipy/blob/master/.github/workflows/tests.yml#L92

PR #1374, opened 2026-02-19
fix eval_configs heads bug

### LINKED ISSUE #1371: eval_config ignores head argument
mace_eval_config() give the same output energy, independent of the --head value
I think the problem originates from utils.py in config_from_atoms() where the head_name property doesn't get updated in accordance with atoms.info['head']

ThomasWarford:
Closes #1371.

I think eval_configs requires some tests. I will put some in #1327 (or another PR) once this is merged.


ilyes319:
thanks @ThomasWarford !!

PR #1392, opened 2026-03-02
add use_edge_irreps_first arg

theochemtheo:
Hello,

In https://github.com/ACEsuit/mace/pull/1267, `use_edge_irreps_first` was added to allow control over the edge irreps in the first interaction block. There currently isn't a way to set this using the training interface.

This PR adds an arg to the training CLI so that this can be set by the user.


alinelena:
@theochemtheo can you retrigger the failed job... is a temp failed download

theochemtheo:
@alinelena I don’t appear to have permissions to trigger the workflows I’m afraid - there’s no restart button for me

PR #1394, opened 2026-03-03
permit first block to be non-linear in 'MACE'-type models

theochemtheo:
Hello,

This PR adds the non-linear interaction block to the list of permitted blocks for the first layer of a `"MACE"` type model.


theochemtheo:
Hiya, anything I can do to unblock this? Cheers!

PR #1420, opened 2026-03-21
Fine tuning with additional arbitrary embeddings. 

williamdavie:
### Fine Tune with additional embeddings.

**1: Foundational model has no joint embeddings**

- All embeddings and head initialized with small weights: ```torch.nn.init.uniform_(param.data, -0.05, 0.05)```
	
**2a: Foundational model has some embeddings**

- weights copied from foundation model *if* specifications are exactly equal.
- otherwise ```torch.nn.init.uniform_(param.data, -0.05, 0.05)```, possibly worth issuing a warning of embedding names are equal but dims are not.

**2b: Head construction**

- before passing the model to MACE class, args.embedding_specs is ordered to match order of any foundation specs found.
- associated weights from foundation model copied *if* out_dims of head equal, otherwise " ".


williamdavie:
there is now a related issue:  #1423

mace/tools/finetuning_utils.py
```diff
@@ -301,6 +339,8 @@ def load_foundations_elements(
     for attr_name, module in model.named_children():
         if attr_name in _handled_attrs:
             continue
+        if not hasattr(model_foundations, "joint_embedding"):
```
aacostadiaz: We could use a more targeted guard here to skip only the missing piece

```
if attr_name not in model_foundations._modules: 
         continue
```

mace/tools/finetuning_utils.py
```diff
@@ -41,11 +41,49 @@ def load_foundations_elements(
...
+                if embedding_spec in foundation_specs.items():
+                    foundation_emb_dim += embedding_spec[1]['emb_dim']
+                    
+            # update head.
+            foundation_emd_head = model_foundations.joint_embedding.project
+            model_emb_head = model.joint_embedding.project
+            head_emb_data = model_emb_head[0].weight.data
+            
+            if foundation_emd_head[0].out_features == model_emb_head[0].out_features:
+                head_emb_data[:,:foundation_emb_dim] = foundation_emd_head[0].weight.data
```
aacostadiaz: I think we might hit a shape error here if the new model drops any foundation embeddings.

`foundation_emb_dim` only counts the embeddings that match exactly between the two models, but the right-hand side contains all of them. If the new model drops or changes any embedding the foundation had the right-hand side will be wider than the slice, raising a shape error.

Could we size the slice based on the foundation head's actual width or check that all foundation embeddings are present before?

aacostadiaz (commented):
Thanks for putting this together! Everything looks mostly good to me. I just left a couple of comments. Happy to approve once we get those sorted out.

mace/tools/train.py
```diff
@@ -765,4 +765,4 @@ def __format__(self, format_spec):
             )
             aux["q95_polarizability"] = compute_q95(delta_polarizability)
 
-        return aux["loss"], aux
+        return aux["loss"], aux
```
aacostadiaz: Nit: it looks like the trailing newline got removed here. Could you add it back?

williamdavie:
Hi @aacostadiaz thanks for commenting on this - apologies for taking some time to review.

Re your requests:

1. I decided to resolve just via the following:

```
head_emb_data[:,:foundation_emb_dim] = foundation_emd_head[0].weight.data[:,:foundation_emb_dim]
```

Given the embeddings are already re-ordered to match the models embeddings and foundation_emd_dim will never exceed the size of the foundation head (by construction), i think this is the best solution. There is, I suppose, some information of the `lost embedding' somewhat encoded into the learned weights of the foundation model - but the effectiveness/influence is more of a scientific question and i think this solution still provides a suitable 'initial guess' at the embedding weights.

2. I have just copied in your suggestion.

3. trivial nit but not exactly sure if i solved it correctly? it should now be exactly as in the current mace version. 

Thanks!

mace/tools/finetuning_utils.py
```diff
@@ -41,11 +41,49 @@ def load_foundations_elements(
...
+                        param.data.copy_(foundation_params[param_name].data)
+                    else:
+                        torch.nn.init.uniform_(param.data, -0.05, 0.05)
+                        
+                # dims for head
+                if embedding_spec in foundation_specs.items():
+                    foundation_emb_dim += embedding_spec[1]['emb_dim']
+                    
+            # update head.
+            foundation_emd_head = model_foundations.joint_embedding.project
```
aacostadiaz, line 144: Thanks, this does fix the crash, but I think it now copies the wrong columns instead of raising. The re-ordering makes the *model's* first `foundation_emb_dim` columns the matching specs, but the *foundation* head's first `foundation_emb_dim` columns are still its own first specs, so as soon as one foundation embedding is dropped or changed the two blocks no longer line up and the copy silently lands the wrong weights. Could you index the foundation side by its own spec offsets rather than slicing from 0?

```python
foundation_specs = {"A": {"type":"categorical","num_classes":3,"emb_dim":8,"per":"node"},
                    "B": {"type":"categorical","num_classes":5,"emb_dim":4,"per":"node"}}
new_specs       = {"B": {"type":"categorical","num_classes":5,"emb_dim":4,"per":"node"}}

# after re-ordering: model specs = ["B"], foundation_emb_dim = 4
# head_emb_data[:, :4] receives columns 0:4 from the foundation head, which belong to "A", not "B"
```

williamdavie: Well spotted thanks, have a look at the most recent commit, I think this now renders the re-ordering useless? Let me know and i can remove that.

aacostadiaz:
Hi @williamdavie, sorry for the slow reply. Short answer to your question: yes, the re-ordering was the actual problem. `head_start` only advances on matching specs, but the columns of `project[0]` span all of the model's specs, so a non-matching spec in front shifts the copy.

I have pushed the fixes to your branch rather than send you round again. Had to merge `develop` in first, since `tests/unit/` did not exist on your branch yet.

Two small things. The model side now builds its slices over every spec, mirroring the foundation side. And I swapped `_modules` for `named_children()`, since the snippet I gave you in July trips pylint and was never going to pass the lint job.

And the important one. Your block places the head columns by name. But right at the end of `load_foundations_elements` there is a loop that copies every foundation parameter whose name and shape match the target, and `joint_embedding.project.0.weight` matches both. So that loop copied the foundation head over yours position by position, undoing what your block had just worked out. Column 0 of the new head is `extra` and column 0 of the foundation head is `dropped`, so `extra` came out holding `dropped`'s weights. That loop predates this PR so it is not something you introduced, but it does mean your block never took effect. It now skips `joint_embedding.`, the way it already skips `readouts.`.

I also added four tests to `tests/unit/test_finetuning_utils.py` for a foundation and target whose specs only partly overlap. Two of them fail on the previous state of the branch, which is what made both problems visible.

Thanks for bearing with the back and forth on this one.

aacostadiaz (approved):

PR #1441, opened 2026-04-14
Add PolarMACE support to MaceTorchSimModel

orionarcher:
## Summary

- Wire PolarMACE-specific inputs (`rcell`, `volume`, `external_field`, `fermi_level`, `density_coefficients`) through the TorchSim interface with sensible defaults matching `AtomicData.from_config()`
- Auto-detect PolarMACE models via `isinstance` check (matching the ASE calculator pattern) and forward additional outputs (`charges`, `dipole`, `density_coefficients`, etc.) in the results dict
- Include padding support for polar tensors in the `torch.compile` path

## Design decisions

- **Auto-detection** via `isinstance(model, PolarMACE)` — zero user burden, matches ASE calculator pattern
- **Defaults** match MACE's own `AtomicData.from_config()`: zeros for `external_field`, `0.0` for `fermi_level`, `zeros(n_atoms, 1)` for `density_coefficients`
- **`external_E_field`** read from TorchSim `SimState` extras (via `SystemExtras.EXTERNAL_E_FIELD`), mapped to `external_field` in MACE's data_dict
- **Cell format unchanged** — both MACE and PolarMACE use `prepare_graph()` with `cell.view(-1, 3, 3)`, so existing `[n_systems, 3, 3]` format works for both

## Test plan

- [x] `test_torchsim_polar_basic` — forward pass with defaults (no extras)
- [x] `test_torchsim_polar_with_extras` — forward pass with explicit extras
- [x] `test_torchsim_polar_no_extras_vs_zero_extras` — verify defaults match zero extras
- [x] `test_torchsim_polar_batched` — batched multi-system forward pass
- [x] `test_torchsim_polar_batched_with_extras` — batched with per-system extras
- [x] `test_torchsim_polar_matches_ase` — compare against ASE `MACECalculator(model_type="PolarMACE")`
- [ ] Existing `test_torchsim.py` tests should pass unchanged (CI verification needed)


orionarcher:
@janosh would you mind taking a look at this? Trying to get MACE Polar working here.

janosh:
Thanks for working on this @orionarcher!

1. `polarizability` and `polarizability_sh` are missing from `per_system_keys`. i think in the `torch.compile` padded path they'll be returned un-sliced, including padding garbage?

2. `_build_polar_data` calls `torch.linalg.inv` unconditionally on the cell, but `AtomicData.from_config()` guards against zero-volume cells (non-periodic systems) by falling back to `zeros(3, 3)`. could error or produce NaN for molecules

3. `density_coefficients` defaults to `zeros(n_atoms, 1)`, hardcoding the second dimension to 1. But the actual dimension depends on the model's `atomic_multipoles_max_l` — e.g. if `max_l=1`, the dimension is `(1+1)^2 = 4`, not 1. The ASE calculator reads this from the model (`self.density_dim = (model.atomic_multipoles_max_l + 1) ** 2`). i think we need do the same here to avoid a shape mismatch when a model with `max_l > 0` is used.

5. `isinstance(model, PolarMACE)` for auto-detection is probably fine for now but could break later this year. there are plans to consolidate MACE model classes. maybe better to detect capability via a model attribute (e.g. `hasattr(model, 'use_polarizability')` or similar flag) rather than relying on the class hierarchy. @ilyes319 would know best

6. `AtomicData.from_config()` defaults `total_spin` to `1.0` while `SimState.spin` defaults to `0.0`. your ASE comparison test handles this correctly, but users won't know to align them. probably worth a log warning when `_is_polar and spin == 0`.

7. `np.random.RandomState(0)` in tests is deprecated in favor of `np.random.default_rng(seed=0)`.

PR #1475, opened 2026-05-14
Fix types.MethodType of cueq_conv_fusion with CueqConvFusionWrapper for LAMMPS compatibility

malixian:
## Summary

This PR replaces the previous runtime `types.MethodType`-based `forward` monkey patch with a standard `torch.nn.Module` wrapper: `CueqConvFusionWrapper`.

## Motivation

Before the fix, `with_cueq_conv_fusion` replaced the `forward` method of the underlying cuEq module dynamically via `types.MethodType`. This worked in the regular Python/ASE execution path, but was not reliably preserved when the model
was exported and loaded through the LAMMPS MLIAP path.

As a result, LAMMPS could call a raw `cuet.SegmentedPolynomial` with the
MACE-style convolution signature:

```python
conv_tp(node_feats, edge_attrs, tp_weights, edge_index)
```

while cuet.SegmentedPolynomial.forward expects:

```python
conv_tp(inputs, input_indices, output_shapes, output_indices)
```

Therefore, node_feats was interpreted as the inputs sequence. For example, for node_feats.shape == [1586, 128], cuEq observed len(inputs) == 1586, instead of the expected 3, which triggered:
```bash
AssertionError: the number of inputs must match the number of inputs of the polynomial
```

### DIFF
--- a/mace/modules/wrapper_ops.py
+++ b/mace/modules/wrapper_ops.py
@@ -130,13 +130,29 @@ def with_scatter_sum(conv_tp: torch.nn.Module) -> torch.nn.Module:
     conv_tp.forward = types.MethodType(forward, conv_tp)
     return conv_tp
 
+class CueqConvFusionWrapper(torch.nn.Module):
+    """A serializable wrapper around cuet.SegmentedPolynomial / mptp.ff.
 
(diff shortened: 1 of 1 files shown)


aacostadiaz (approved):
Thanks for fixing this!

PR #1477, opened 2026-05-14
Fix dataset content logging to count labels, not sum weights

vue1999:
`log_dataset_contents` reported energy/force/etc. counts by summing `config.property_weights[prop_name]` across the dataset. This relied on weights being 0.0 (missing) or 1.0 (present), and produced inflated counts whenever users set `config_<name>_weight `to a non-unit value. Replaced with a direct count of configs where the label is present (`properties[prop_name] is not None`), so the logged numbers match the actual number of labels in train/valid/test sets.


WillBaldwin0:
yep, this is a good catch!

PR #1478, opened 2026-05-15
LES extension (dipole, quadrupoles, polarizability, etc.) merge clean

djkim-kr:
## Add MACELES: long-range electrostatics implementation with added features (dipoles, quadrupoles, polarizability, etc.)

> Hi,
> 
> This is Dongjin Kim from UC Berkeley working with Prof. Bingqing Cheng.
> Here's a new updated version of our LES implementation for MACE.
> While the extended LES method is well described in our preprint: [https://arxiv.org/abs/2605.05746](https://arxiv.org/abs/2605.05746), I briefly describe the main changes in this update below.
> 
> Now, this branch is mergeable with your most recent `main` and is fully backward compatible with our `main` branch. Please check and make changes if needed.
> 
> Thank you very much for your consideration.
> Best,
> Dongjin

### `mace/modules/extensions.py` — MACELES class

**`class MACELES` (line 123)**

MACELES augments the model wiht per-atom electrostatic
degrees of freedom predicted from the equivariant node features and
passing them to the `les` library to calculate long-range part.

This PR adds extended features including dipoles, quadrupoles, polarizability, etc.

**Readout architecture (`__init__`, lines 124–226)**

Each quantity uses a dedicated readout module cloned from the MACE readout
stack via two helper functions:

- `_copy_mace_readout(readout, change_irrep_out=...)` (line 50): extended
  from upstream's version with an optional `change_irrep_out` argument.
  Used to clone the existing readout and retarget its output irreps —
  e.g. `change_irrep_out="1x1o"` for dipoles, `"1x2e"` for quadrupoles.

- `_copy_mace_readout_tp(readout, ...)` (line 77, new): builds a
  `LinearLesReadoutBlock` or `NonLinearLesReadoutBlock` for quantities
  that require an outer product of l=1 features (see `blocks.py` below).

The per-quantity readouts are:

| Quantity | Approach | Irreps |
|---|---|---|
| Atomic charges (`les_readouts`) | scalar readout, cloned from MACE readout | `0e` |
| Atomic dipoles (`les_u_readouts`) | cloned with `change_irrep_out="1x1o"` | `1o` |
| Quadrupoles via l=2 (`les_quad_2e_readouts`) | cloned with `change_irrep_out="1x2e"`, then converted via `change_of_basis_quads` (`ReducedTensorProducts('ij=ji', i='1o')`) to Cartesian 3×3 | `2e` |
| Quadrupoles via l=1 (`les_quad_1o_readouts`) | outer product v⊗v using `_copy_mace_readout_tp` → `LinearLesReadoutBlock` | `1o ⊗ 1o` |
| Isotropic polarizability (`les_alpha_readouts`) | scalar readout | `0e` |
| Anisotropic polarizability via l=2 (`les_alpha_2e_readouts`) | cloned with `change_irrep_out="1x0e + 1x2e"`, converted via `CartesianTensor("ij=ji").reduced_tensor_products().change_of_basis` | `0e + 2e` |
| Anisotropic polarizability via l=1 (`les_alpha_1o_readouts`) | outer product v⊗v using `_copy_mace_readout_tp` → `LinearLesReadoutBlock` | `1o ⊗ 1o` |
| Screening charges κ (`les_kappa_readouts`) | scalar readout | `0e` |

The l=1 outer-product path (`les_quad_1o_readouts`, `les_alpha_1o_readouts`)
exists because many trained MACE models use only l≤1 hidden irreps and
therefore have no `2e` features — the quadrupole/polarizability tensor
must then be formed as v⊗v from the available `1o` channel.

**`keep_last_layer_irreps` requirement (line 126)**

```python
kwargs["keep_last_layer_irreps"] = True
```

MACELES forces this flag before calling `super().__init__()`. Without it,
`MACE.__init__` would collapse the last interaction layer to scalar-only
output (see `models.py` fix below), making vector/tensor features
unavailable at readout time.

---

### `mace/modules/blocks.py` — LES readout blocks (lines 1400–1555)

**`LinearLesReadoutBlock` (line 1400)**

Predicts a per-atom 3×3 symmetric polarizability tensor from equivariant
features. For each node:

1. Apply an equivariant `Linear` layer mixing features within the input
   irreps.
2. Extract scalar (`0e`) weights αᵢₖ and vector (`1o`/`1e`) channels vᵢₖ.
3. Map scalar weights to a mixing matrix via a learned linear
   `scalar_to_weight_1o` / `scalar_to_weight_1e` layer.
4. Form the outer product: `T = Σₖ wₖ · (vₖ ⊗ vₖ)` — this is
   automatically symmetric and equivariant (SO(3)-covariant under
   rotations of the input features).
5. Optionally enforce positive-definite trace via softplus on the weights
   (`make_w_pos=True`).

The outer product lives entirely in standard PyTorch (no e3nn CG
coefficients needed for the contraction), keeping compilation overhead low.

**`NonLinearLesReadoutBlock` (line 1502)**

Nonlinear variant: passes the equivariant features through a gated
nonlinearity (`e3nn.nn.Gate`) before the same outer-product readout.
Used when the linear model undershoots the polarizability dynamic range.

---

### `mace/modules/models.py` (line 169)

```python
# before
if num_interactions == 1:
    hidden_irreps_out = str(hidden_irreps[0])

# after
if num_interactions == 1 and not keep_last_layer_irreps:
    hidden_irreps_out = str(hidden_irreps[0])
```

When `num_interactions=1`, upstream unconditionally reduced the hidden
irreps to scalars only, even when `keep_last_layer_irreps=True`. MACELES
sets this flag to preserve vector/tensor features; without the fix,
`hidden_irreps_out` becomes e.g. `"128x0e"` instead of the full irreps,
and the downstream LES readout blocks raise error due to lack of L=1 or L=2 node features.

---

### `mace/data/atomic_data.py` — `atomic_numbers` field (line 402)

```python
cls_kwargs = dict(
    ...
+   atomic_numbers=torch.tensor(config.atomic_numbers, dtype=torch.long),
    ...
)
```

Required for future use of element-specific fixed charge options in the LES
library (e.g. `use_fixed_atomic_charges`).
`AtomicData.from_config` was not populating this field, causing a
`KeyError` when those LES options are active.

---

### `mace/calculators/mace.py` — BEC forces under external electric fields and LES outputs

**`MACECalculator.__init__` (line ~112)**

Five new parameters: `compute_bec`, `external_field`, `eps_infty`,
`electric_field_unit`, `keep_neutral`. When `compute_bec=True`, `"bec"`
is added to `implemented_properties`.

**`MACECalculator.calculate` — forward pass (lines 644–683)**

`compute_bec=True` is forwarded via `forward_kwargs`.

The LES latent outputs (`latent_alphas`, `latent_kappas`, `BEC`) are
collected into Python lists (`ret_tensors.setdefault(..., []).append(...)`),
so they are accessible as ASE calculator results during MD.

**`MACECalculator.calculate` — post-processing (lines 760–803)**

After the `results_map` block, the list-accumulated LES quantities are
averaged over ensemble models and stored:

```python
self.results["LES_alphas"] = torch.mean(torch.stack(...), dim=0).cpu().numpy()
self.results["LES_kappas"] = torch.mean(torch.stack(...), dim=0).cpu().numpy()
self.results["bec"]        = torch.mean(torch.stack(...), dim=0).cpu().numpy()
```

When `external_field` is set, atomic forces are augmented by the BEC
contribution:
```
F_i  +=  Z*_i · E_eff
E_eff = E_ext · sqrt(ε_r)
ε_r  = ε_∞ / (1 + χ),   χ = Σᵢ α_i / (V ε₀)
```
`keep_neutral=True` subtracts the mean BEC before the force application
to enforce acoustic sum rule neutrality.
More details are described in our preprint.

---

### `mace/cli/eval_configs.py` — latent quantity output

`run()`: added collection lists `us_list`, `kappas_list`, `alphas_list`
alongside the existing `qs_list`. Per-structure latent quantities are
split from the batch using `batch.ptr` and written to `atoms.arrays` as
`{prefix}latent_dipoles`, `{prefix}latent_kappas`, `{prefix}latent_alphas`.
This enables inspection of the internal LES state without modifying the
model forward pass.

---

### Backward compatibility

With this branch, training scripts and pre-trained `MACELES` potentials from `ChengUCB/mace:main` load and run without modification. (The training scripts and potentials are 
shared in [https://github.com/ChengUCB/extended_les_fit/tree/main/MLIPs/MACE-LES](https://github.com/ChengUCB/extended_les_fit/tree/main/MLIPs/MACE-LES))

`MACELES.forward` wraps attribute accesses that were
added in later versions (`les_output_scale`, `les_kappa_scale`,
`les_alpha_scale`, `use_anisotropic_polarizability`) with `hasattr` guards
to handle older checkpoints.


mace/modules/blocks.py
```diff
@@ -1395,3 +1395,145 @@ def __repr__(self):
             else f"{self.shift.item():.4f}"
         )
         return f"{self.__class__.__name__}(scale={formatted_scale}, shift={formatted_shift})"
+
+# LES-specific readout blocks
+class LinearLesReadoutBlock(torch.nn.Module):
```
aacostadiaz, line 1401: Could we add tests for the new outputs? A rotation/equivariance check on the new tensor readouts would be ideal

mace/modules/blocks.py
```diff
@@ -1395,3 +1395,145 @@ def __repr__(self):
...
+        self.scalar_slices: List[Tuple[int, int]] = []
+        self.vector_1o_slices: List[Tuple[int, int]] = []
+        self.vector_1e_slices: List[Tuple[int, int]] = []
+
+        n_scalar = 0
+        n_1o = 0
+        n_1e = 0
+        offset = 0
+
+        for mul, ir in self.irreps_in:
```
aacostadiaz, line 1435: With cueq the features are in ir_mul layout not mul_ir, so this reshape mixes up the vector components

mace/calculators/mace.py
```diff
@@ -649,6 +671,13 @@ def calculate(self, atoms=None, properties=None, system_changes=all_changes):
                 if out.get(key) is not None:
                     val[i] = out[key].detach()
 
+            if "latent_alphas" in out and out["latent_alphas"] is not None:
+                ret_tensors.setdefault("latent_alphas", []).append(out["latent_alphas"].detach())
```
aacostadiaz: I think this breaks with more than one model. These are python lists and the `val[i] = ...` loop above does `list[1] = ...` on the second model we will get indexerror. Preallocate them like the other outputs.

mace/calculators/mace.py
```diff
@@ -731,6 +760,47 @@ def calculate(self, atoms=None, properties=None, system_changes=all_changes):
...
+        if "latent_kappas" in ret_tensors:
+            self.results["LES_kappas"] = torch.mean(torch.stack(ret_tensors["latent_kappas"]), dim=0).cpu().numpy()
+        if getattr(self, "compute_bec", False) and "bec" in ret_tensors:
+            self.results["bec"] = (
+                torch.mean(torch.stack(ret_tensors["bec"]), dim=0).cpu().numpy()
+            )
+        if self.external_field is not None and getattr(self, "compute_bec", False) and "bec" in self.results:
+            bec_output = self.results["bec"]  # [N_atoms, 2, 3, 3] or [N_atoms, 3, 3]
+            if bec_output.ndim == 4:
+                if bec_output.shape[0] == 2:
```
aacostadiaz: This looks fragile, for a 2-atom system shape[0] == 2 matches the atom axis and we sum over atoms. From what I can see les always stacks the contributions on dim 1, so isn't it simpler to just sum over dim 1 when ndim == 4?

mace/modules/blocks.py
```diff
@@ -1395,3 +1395,145 @@ def __repr__(self):
...
+
+    def __init__(
+        self,
+        irreps_in: o3.Irreps,
+        hidden_dim: int = 32,
+        cueq_config: Optional[CuEquivarianceConfig] = None,
+    ):
+        super().__init__()
+        self.irreps_in = irreps_in
+        self.cdim = int(str(irreps_in).split("x")[0])
```
aacostadiaz: This assumes all irreps have the same multiplicity and come ordered [0e, 1o, ...]. Instead of hard-coding cdim, loop over irreps_in and record the (start, end) of each 0e and 1o block, like LinearLesReadoutBlock does in its init.

aacostadiaz (commented):
Thanks for the work here! Two things before merge: there are no tests yet, and I think the new tensor outputs give wrong numbers on the cueq backend.

gabor1:
@cursor review

djkim-kr:
## Changes addressing review feedback

Thank you for the prompt review!
I've committed a few changes to the existing code.
I hope these changes address the review comments on the LES tensor readouts, calculator, and tests for new outputs.

**`mace/calculators/mace.py`**
- Fixed an IndexError with multi-model committees: the LES outputs (`latent_alphas`, `latent_kappas`, `BEC`) are now preallocated per model instead of collected in Python lists.
- Fixed BEC force handling for small systems: a 4-D BEC is now summed over axis 1 (where `les` stacks the charge/dipole contributions).

**`mace/modules/blocks.py`**
- `LinearLesReadoutBlock` / `NonLinearLesReadoutBlock`: handle the cueq `ir_mul` layout correctly. Vectors are now reshaped according to the active layout, so the 1o components are no longer mixed up under cueq. The `mul_ir (e3nn) path is unchanged.
- `NonLinearLesReadoutBlock`: stop hard-coding `cdim` from the irreps string (which assumed uniform multiplicity and `[0e, 1o, ...]` ordering). It now records the 0e/1o block slices by iterating `irreps_in`, mirroring `LinearLesReadoutBlock`.
- Both blocks lazily backfill the new attributes in `forward`, so checkpoints saved before this change still load and run (backward compatible).

**`tests/test_maceles.py`**
- Add `test_les_readout_equivariance`: checks the predicted 3x3 tensor transforms as `R T R^T` under random input rotations (SO(3)-equivariance), for both readout blocks.
- Add `test_les_readout_layout_invariance`: checks the `ir_mul` path recovers the same vectors as `mul_ir`, using non-uniform irreps to also cover the slice-based multiplicity handling.
- Update `test_run_eval_no_bec` to reflect that latent charges are always inferred by MACELES (independent of `--compute_bec`).

+++
Cleaned up `blocks.py` and `extensions.py` to pass the pre-commit checks
(isort + pylint) e.g. removed unused imports, dropped an unused loop
variable, and simplified `if/else` loop. No functional changes.
I think it should pass the pre-commit test without a problem.

mace/modules/blocks.py
djkim-kr: Thank you. We now added the test for the new tensor outputs.

mace/modules/blocks.py
djkim-kr: Thank you. We didn't realize that they used a different layout. Now modified the code to deal with `cueq` as well without problem.

mace/calculators/mace.py
djkim-kr: Thank you. Now we preallocate them like the ohter outputs

mace/calculators/mace.py
djkim-kr: True. We followed your suggestion. Thank you.

mace/modules/blocks.py
djkim-kr: Thank you. Now it is more flexible as we loop over irreps_in and record the (start, end) of each 0e and 1o block instead of hard-coding them.

mace/calculators/mace.py
```diff
@@ -731,6 +762,44 @@ def calculate(self, atoms=None, properties=None, system_changes=all_changes):
...
+                epsilon_r = self.eps_infty / (1.0 + chi)
+            else:
+                epsilon_r = getattr(self, "eps_infty", 1.0) if getattr(self, "eps_infty", None) is not None else 1.0
+            e_ext_arr = np.array(self.external_field, dtype=np.float64)
+            scaled_e_field = e_ext_arr * (epsilon_r ** 0.5)
+            electric_field_unit = getattr(self, "electric_field_unit", 1.0)
+            if np.isscalar(scaled_e_field) or scaled_e_field.ndim == 0:
+                forces_bec = bec_output * scaled_e_field * electric_field_unit
+            else:
+                forces_bec = bec_output @ scaled_e_field * electric_field_unit
```
aacostadiaz: I think this contracts the wrong axis. In les, bec.py builds the BEC as

    #Grad returns P on 2nd index, BEC is defined with P on first index
    bec_basis_complex = grad(y=P, x=r).transpose(1,2).contiguous()

so the axes are [atom, polarization, displacement] and the field force is F_nj = sum_i Z*_nij E_i. `bec_output @ scaled_e_field` contracts the last axis (displacement) instead, which gives the transposed result.

```suggestion
                forces_bec = np.einsum("nij,i->nj", bec_output, scaled_e_field) * electric_field_unit
```

mace/calculators/mace.py
```diff
@@ -731,6 +762,44 @@ def calculate(self, atoms=None, properties=None, system_changes=all_changes):
...
+                else:
+                    chi = 0.0
+                epsilon_r = self.eps_infty / (1.0 + chi)
+            else:
+                epsilon_r = getattr(self, "eps_infty", 1.0) if getattr(self, "eps_infty", None) is not None else 1.0
+            e_ext_arr = np.array(self.external_field, dtype=np.float64)
+            scaled_e_field = e_ext_arr * (epsilon_r ** 0.5)
+            electric_field_unit = getattr(self, "electric_field_unit", 1.0)
+            if np.isscalar(scaled_e_field) or scaled_e_field.ndim == 0:
+                forces_bec = bec_output * scaled_e_field * electric_field_unit
```
aacostadiaz: I tried passing a single number as external_field and it fell over here: the result stays [N,3,3] and won't add to the forces [N,3]. Should this branch be there at all?

aacostadiaz (commented):
Thanks for addressing the review. I found some issues in the external-field force path, details inline.

mace/calculators/mace.py
djkim-kr: Thanks for catching this! You're absolutely right. We had updated the BEC
output in `les` to follow the conventional definition (Z*_ij = ∂P_i/∂r_j,
polarization on the first index), but the external-field force path here was
still using the old contraction (`bec @ E`), which contracts the displacement
axis instead of the polarization axis.

Since the symmetric part of Z* contributes either way identically, the error
only shows up in the antisymmetric (off-diagonal) part, which is small for
our systems (the BEC is nearly symmetric), so the impact on actual simulations
is minor. But it's a genuine bug, and your suggested `np.einsum("nij,i->nj", ...)`
is correct. Fixed and committed. Thanks!

mace/calculators/mace.py
djkim-kr: Agreed, this shouldn't be here. I guess it was leftover from some earlier testing and
never cleaned up. A field is a 3-vector, and the scalar path produced an
[N, 3, 3] array that couldn't be added to the forces anyway. Removed. We also
normalize `external_field` to a (3,) vector in `__init__` now (with a clear
error otherwise), so both the model and force paths stay consistent. Thanks!

aacostadiaz (approved):
Thanks, both fixes look right.

PR #1501, opened 2026-06-13
Fix non-PBC fake-cell sizing in get_neighborhood

keceli:
The previous formula sized the fake cell as `max(|positions|) * 5 * cutoff`, which depended on where the molecule sat relative to the origin rather than its actual extent, and used a multiplicative safety margin instead of an additive one. A molecule with max|positions|=237 A and cutoff=5 produced a fake cell of ~5948 A per side. matscipy's neighbour_list then allocated a linked-cell grid of ~1.7 billion bins (~6 GiB CPU RAM) just to find a few hundred edges among ~50 atoms. In production this propagated into multi-hundred-GiB tensor allocations on accelerators and triggered uncatchable GPU page faults that took down 12 000-rank MPI runs.

Replace with cell = extent + 5 * cutoff so the cell scales with the molecular bounding box and is translation-invariant. For a 200 A molecule the cell drops from ~6000 A to ~232 A; for a typical 10 A molecule from 250 A to ~35 A. Edge counts are unchanged; matscipy allocation goes from O(GiB) to O(MiB).

PS: I had out of memory errors for some geometries and opus4.7 found the bug and helped me patch.

### DIFF
--- a/mace/data/neighborhood.py
+++ b/mace/data/neighborhood.py
@@ -24,16 +24,23 @@ def get_neighborhood(
     pbc_y = pbc[1]
     pbc_z = pbc[2]
     identity = np.identity(3, dtype=float)
-    max_positions = np.max(np.absolute(positions)) + 1
-    # Extend cell in non-periodic directions
-    # For models with more than 5 layers, the multiplicative constant needs to be increased.
-    # temp_cell = np.copy(cell)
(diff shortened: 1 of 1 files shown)


gabor1:
this fixes the same issue as #1473 . are you happy with the fix there?

keceli:
Oh, yes, didn't see it. Thank you!

PR #1505, opened 2026-06-22
Make node_energy Mac Silicon MPS-compatible (avoid hardcoded float64)

aakolganov:
## Issue

The Apple-Silicon GPU MPS backend does not support float64, so this raises at inference on `device="mps"`, even when the model is loaded/run in float32:

**TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.**

## Fix

Keep the float64 on CPU/CUDA (unchanged behaviour) and fall back to the tensor's working dtype only on MPS:

```python
energy_dtype = node_e0.dtype if node_e0.device.type == "mps" else torch.float64
node_energy = node_e0.clone().to(energy_dtype) + node_inter_es.clone().to(energy_dtype)

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Low Risk**
> Single-line behavioral change in ScaleShiftMACE forward; MPS gets float32 node_energy instead of crashing, CPU/CUDA behavior unchanged.
> 
> **Overview**
> Fixes **ScaleShiftMACE** inference on Apple Silicon when tensors live on **MPS**, where PyTorch cannot use float64.
> 
> **`node_energy`** was built by cloning **`node_e0`** and **`node_inter_es`** with **`.double()`**, which crashes on MPS. The forward pass now picks **`energy_dtype`**: **`node_e0.dtype`** on MPS (typically float32) and **`torch.float64`** on CPU/CUDA, then sums the clones in that dtype. **`total_energy`**, forces, and other outputs are unchanged; only the per-atom energy tensor’s accumulation dtype is device-aware.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7117d352cccd4c6d7fba5e15777fbc40be66cdf4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/mace/modules/extensions.py
+++ b/mace/modules/extensions.py
@@ -23,7 +23,12 @@ from mace.modules.blocks import (
     NonLinearReadoutBlock,
 )
 from mace.modules.models import ScaleShiftMACE
-from mace.modules.utils import get_atomic_virials_stresses, get_outputs, prepare_graph
+from mace.modules.utils import (
+    get_atomic_virials_stresses,
+    get_outputs,
+    prepare_graph,
+    safe_double,
+)
 from mace.modules.wrapper_ops import (
     CuEquivarianceConfig,
     OEQConfig,
@@ -237,7 +242,7 @@ class MACELES(ScaleShiftMACE):
         inter_e = scatter_sum(node_inter_es, data["batch"], dim=-1, dim_size=num_graphs)
 
         total_energy = e0 + inter_e
-        node_energy = node_e0.clone().double() + node_inter_es.clone().double()
+        node_energy = safe_double(node_e0.clone()) + safe_double(node_inter_es.clone())
 
         les_q = torch.sum(torch.stack(node_qs_list, dim=1), dim=1)
         les_result = self.les(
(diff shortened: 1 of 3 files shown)


ilyes319:
@cursor review

aacostadiaz:
Thanks @aakolganov for tracking this down and for the fix, much appreciated!

I've pushed a small follow-up commit on top of yours that generalizes it: the same hardcoded `.double()` pattern exists in `extensions.py`, so those paths would still raise the same TypeError on MPS. Behaviour on CPU/CUDA is unchanged.

Verified on Apple Silicon: ScaleShiftMACE and PolarMACE forwards run on mps in float32, and MACE-MP-small through the ASE calculator gives energies/forces matching CPU float32 to ~1e-7.

One heads-up, out of scope for this PR: `MACECalculator` still fails earlier on MPS when loading a float64 checkpoint, because `torch.load(..., map_location=device)` (mace/calculators/mace.py) can't materialize float64 weights on mps. Workaround for now is torch.load(path, map_location="cpu").float().to("mps");

aacostadiaz (approved):

PR #1522, opened 2026-07-13
removing ipex, and fixing multigpu detection help from @AbrarFaiyad

alvarovm:
We propose to remove IPEX modules for Intel XPU training and ASE calculator. IPEX was retired ([archived](https://github.com/intel/intel-extension-for-pytorch)) by Intel in favor of XPU support within Pytorch >=2.8.
We also fixed the support for multigpu and multinode training on large Intel GPU clusters.

---

> [!NOTE]
> **Medium Risk**
> Touches distributed training device mapping on XPU clusters; behavior change for users who still relied on IPEX, but scope is limited to Intel XPU paths.
> 
> **Overview**
> **Removes Intel Extension for PyTorch (IPEX)** from the ASE `MACECalculator` and `run_train.py`, including all `ipex.optimize` paths, in favor of native `torch.xpu` support in recent PyTorch.
> 
> **XPU startup** no longer requires importing IPEX/oneCCL; it only logs when `--device=xpu` is set but `torch.xpu` is missing. **Distributed XPU** now clamps `torch.xpu.set_device` when `local_rank` exceeds the visible device count, and **DDP** uses `device_ids=[0]` when each process sees a single XPU (e.g. under `ZE_AFFINITY_MASK`), avoiding overflow errors from invalid `local_rank` device IDs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc9674a8dfaf1310cec5c53f7e4e2265abae9370. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ilyes319:
@cursor review

mace/cli/run_train.py
```diff
@@ -97,13 +97,9 @@ def run(args) -> None:
...
     if args.device == "xpu":
-        try:
-            import intel_extension_for_pytorch as ipex
-            import oneccl_bindings_for_pytorch as oneccl  # pylint: disable=unused-import
-        except ImportError as e:
-            raise ImportError(
-                "Error: Intel extension for PyTorch not found, but XPU device was specified"
-            ) from e
+        if not hasattr(torch, "xpu"):
+            logging.info(f"Device GPU: XPU requested but torch.xpu not found, cannot use --device=xpu")
```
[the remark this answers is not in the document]
alvarovm: Updated removed since devices are checked at init_device()

mace/cli/run_train.py
```diff
@@ -930,14 +938,6 @@ def run(args) -> None:
...
         return
 
-    if args.device == "xpu":
-        try:
-            model, optimizer = ipex.optimize(model, optimizer=optimizer)
-        except ImportError as e:
-            logging.error(
-                "Intel Extension for PyTorch not found, but XPU device was specified. "
-                "Please install it to use XPU device."
-            )
```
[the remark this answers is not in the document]
alvarovm: This is fine. Here `model` is not distributed.

mace/cli/run_train.py
```diff
@@ -893,7 +890,18 @@ def run(args) -> None:
...
+        # raise "value cannot be converted to type int without overflow".
+        # Pin DDP to device 0 in that case.
+        _ddp_dev = local_rank
+        if args.device == "xpu":
+            try:
+                if torch.xpu.device_count() == 1:
+                    _ddp_dev = 0
+            except Exception:
+                _ddp_dev = 0
+        distributed_model = DDP(model, device_ids=[_ddp_dev])
```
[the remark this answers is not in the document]
alvarovm: I'd say is ok. One usually runs 1 mpi rank  for each device

alvarovm (commented):
Removed useless warnings. Reviewed code.

mace/cli/run_train.py
aacostadiaz: There's a second DDP here in the final evaluation loop that still uses the raw local_rank. I think this is the line the bot was actually referring to in https://github.com/ACEsuit/mace/pull/1522/changes/BASE..bc9674a8dfaf1310cec5c53f7e4e2265abae9370#r3574263042

```python
distributed_model = DDP(model, device_ids=[_ddp_dev])
```

mace/cli/run_train.py
```diff
@@ -97,13 +97,9 @@ def run(args) -> None:
     update_keyspec_from_kwargs(args.key_specification, vars(args))
 
     if args.device == "xpu":
-        try:
```
aacostadiaz, line 101: I think removing this import breaks distributed XPU. Importing `oneccl_bindings_for_pytorch` was what registered the "ccl" backend and `distributed_tools.py` still calls `init_process_group(backend="ccl")` for xpu

aacostadiaz (commented):
Thanks for this cleanup! I've left a few comments below. 

Since we don't have Intel hardware in CI, we can't test any of the XPU paths ourselves. Could you re-run a multi-node job on your cluster once the comments are addressed and confirm it works end to end?

mace/cli/run_train.py
alvarovm: The `oneccl_bindings_for_pytorch`  is also retired per intel see [https://github.com/intel/torch-ccl](https://github.com/intel/torch-ccl)

mace/cli/run_train.py
alvarovm: torch >2.8 DDP supports Intel CCL back-end.

mace/cli/run_train.py
aacostadiaz: Right, I think that means this PR needs one more change to go with it, in `distributed_tools.py`

Line 56 still asks for backend="ccl" for xpu, and that name was only ever registered as a side effect of importing `oneccl_bindings_for_pytorch`. Now that the import is gone, nothing registers it. 

So as the PR stands, distributed XPU runs fail at process-group init. Could you switch that call to the native XPU backend? torch's own docs say "the known default registrations today are: nccl for cuda, gloo for cpu, xccl for xpu"

```python
elif args.device == "xpu":
    torch.distributed.init_process_group(backend="xccl", ... )
```

mace/cli/run_train.py
aacostadiaz: I think we're talking about different things here. The backend isn't the issue, it's the device index. This line is still `DDP(model, device_ids=[local_rank])` with the raw local_rank, while the DDP up in the training loop now uses `_ddp_dev` (line 904). 

On the exact layout this PR fixes on lines 893-904 (ZE_AFFINITY_MASK, local_rank >= visible devices) this one still raises the "value cannot be converted to type int without overflow" error, just at the very end of the run, after all the training is done.

mace/cli/run_train.py
alvarovm: Ok. Added xccl according to [here](https://github.com/pytorch/pytorch/blob/v2.8.0/torch/distributed/distributed_c10d.py#L236).

mace/cli/run_train.py
alvarovm: @aacostadiaz  I see. I modified the lines around DDP to pass the correct rank

mace/cli/run_train.py
```diff
@@ -1049,11 +1045,18 @@ def run(args) -> None:
         )
         model.to(device)
         if args.distributed:
-            # re-enable gradients for distributed model for evaluation of stage-two model
-            # after param.requires_grad = False was called before evaluating stage-one model
-            for param in model.parameters():
-                param.requires_grad = True
```
aacostadiaz, line 1162: I think removing this will reintroduce the bug commit caf60ab0 fixed ("Fix bug that prevents evaluating stage-two model in multi-gpu environment"). Could you add it back?

alvarovm: I reverted this change

aacostadiaz:
LGTM! 

Pushed a small commit to fix the failing lint job

PR #1523, opened 2026-07-14
Fix stress normalization for partially periodic systems (#1509)

### LINKED ISSUE #1509: Analytic stress on partially periodic (slab) systems is ~18x smaller than the Hellmann-Feynman strain derivative of the energy
**Describe the bug**

For a 2D slab (`pbc=[True, True, False]`, vacuum along z) the ASE calculator's analytic `get_stress()` is inconsistent with the finite-difference strain derivative of the calculator's own potential energy: on a phosphorene monolayer we measure

```
sigma_yy (finite difference dE/d_eps / V) :  3.217e-03 eV/A^3
sigma_yy (analytic get_stress)            :  1.807e-04 eV/A^3
ratio FD / analytic                       :  17.81
```

A fully periodic control with the **same calculator and the same script** (bulk Si, `pbc=[True, True, True]`) is consistent to 4 decimal places (ratio 1.0000). That isolates the inconsistency to the partially periodic geometry rather than the model or the FD setup.

Because the energy surface itself is fine, every energy-derived quantity is unaffected, but everything that consumes the analytic virial silently breaks on slabs (see Additional context).

**To Reproduce**

Run the self-contained script below (CPU, plain e3nn, float64, no cuEquivariance; ~2 min):

```python
import numpy as np
from ase import Atoms
from mace.calculators import mace_mp

calc = mace_mp(model="medium-omat-0", device="cpu", default_dtype="float64",
               enable_cueq=False, dispersion=False)

def phosphorene(vacuum=20.0):
    a, b, dz = 4.62, 3.30, 2.14   # approximate monolayer black phosphorene
    z0 = vacuum / 2
    return Atoms("P4",
                 positions=[(0.000 * a, 0.0, z0), (0.367 * a, b / 2, z0),
                            (0.500 * a, b / 2, z0 + dz), (0.867 * a, 0.0, z0 + dz)],
                 cell=[a, b, vacuum + dz], pbc=[True, True, False])

base = phosphorene().repeat((2, 3, 1))
cell0 = base.get_cell().array.copy()
V = abs(np.linalg.det(cell0))

EPS, DELTA = 0.010, 0.001
def at_eps(e):                      # affine strain along yy, clamped ions
    a = base.copy()
    M = np.eye(3); M[1, 1] += e
    a.set_cell(cell0 @ M, scale_atoms=True)
    a.calc = calc
    return a

energies = {e: at_eps(e).get_potential_energy()
            for e in (EPS - DELTA, EPS, EPS + DELTA)}
sigma = at_eps(EPS).get_stress(voigt=True)

fd = (energies[EPS + DELTA] - energies[EPS - DELTA]) / (2 * DELTA) / V
print(f"sigma_yy finite-diff : {fd: .6e} eV/A^3")
print(f"sigma_yy analytic    : {sigma[1]: .6e} eV/A^3")
print(f"ratio FD/analytic    : {fd / sigma[1]:.4f}")
# observed: ratio ~ 17.8  (expected ~ 1.0)
# control: replace the slab with ase.build.bulk("Si","diamond",a=5.43).repeat((2,2,2))
#          -> ratio 1.0000 with the same calculator
```

**Expected behavior**

At any configuration, sigma_ij = (1/V) dE/d_eps_ij under affine scaling of the positions, so the analytic stress and the finite-difference derivative must agree regardless of model quality (ratio ~1.0, as the fully periodic bulk-Si control gives). On the slab they disagree by ~18x.

**Screenshots**

Not applicable; the text output above shows the discrepancy.

**Environment (please complete the following information):**

 - OS: Linux x86_64 (Ubuntu 24.04)
 - mace-torch 0.3.16, e3nn 0.4.4, ase 3.28.0, numpy 2.4.6
 - torch 2.11.0+cu128, though the repro runs on CPU with plain e3nn at float64 (no cuEquivariance involved)
 - model: `mace_mp(model="medium-omat-0")`; we expect the effect to be model-independent (it sits at the calculator level) and can re-test other models on request

**Additional context**

- The ratio is approximately, but maybe not exactly, constant: independent strain-direction fits on the same slab give 17.8-17.9 (xx vs yy), which may point to a missing/partial virial contribution rather than a pure volume-normalization slip. We did not attempt to localize the mechanism.
- Downstream effects we measured before finding the root cause:
  - stress-slope elastic constants come out ~18x too small, while energy-curvature fits on the same relaxed configurations give values in the expected range (C11 ~ 33, C22 ~ 105 N/m for the monolayer model);
  - `Inhomogeneous_NPTBerendsen` (z-masked) runs away: the barostat balances the *unscaled* kinetic pressure against the ~18x-undersized virial, so the in-plane cell inflates monotonically (~+20 % over 50 ps at 300 K), identically for e3nn/float64, e3nn/float32 and cuEq/float32, so the failure is independent of precision and backend.
- Possibly related (but, as far as we can tell, distinct) reports: #980 (per-atom stress), #294/#222/#395 (training-data virial formats).

Happy to provide the full measurement data (the repro above is extracted from a benchmarking study of MACE+cuEquivariance on consumer GPUs) or to test candidate fixes.

darthjaja6:
### Summary

Fixes #1509. The analytic stress returned for partially periodic (slab) systems was silently scaled down by a large factor (~18x in the reported case), while energies and forces were unaffected.

### Root cause

`get_neighborhood` extends the cell along non-periodic directions so that `matscipy` can bin the atoms:

```python
if not pbc_x:
    cell[0, :] = max_positions * 5 * cutoff * identity[0, :]
```

This inflated cell was then **returned** from the function and stored in `AtomicData`. Stress is normalized by `det(cell)` in `compute_forces_virials`, so the artificial volume along the vacuum direction shrinks the stress by exactly the ratio of the inflated extent to the true extent. For the phosphorene example in the issue this ratio is `max_positions * 5 * cutoff / z_true ≈ 17.8`, matching the reported `17.81`. Energies and forces don't touch `det(cell)`, which is why only stress was affected and the bug went unnoticed.

### Fix

Use the extended cell only for the neighbour-list call, and return the physical cell whenever at least one direction is periodic. `unit_shifts` are zero along non-periodic directions, so the shifts (and therefore the energy/forces) are identical. Fully non-periodic systems keep the extended cell, since stress is meaningless there and long-range models rely on a non-degenerate cell — so their behaviour is unchanged.

### Verification

Finite-difference check (analytic stress vs. strain derivative of the energy), model-independent:

| system | before | after |
|---|---|---|
| bulk Si (pbc TTT, control) | ratio 1.0000 | ratio 1.0000 |
| Si slab (pbc TTF) | ratio ~16.6 | ratio 1.0000 |

Added a regression test (`test_stress_partial_pbc`) that fails on the current code and passes with the fix, plus an assertion in `test_half_periodic` that the physical cell is returned. Full existing test suite passes.

---

> [!NOTE]
> **Medium Risk**
> Touches core neighbourhood/cell plumbing used by all models; behaviour change is scoped to partial PBC (fully periodic and fully aperiodic paths are preserved) but stress and any volume-derived quantities now differ for slabs.
> 
> **Overview**
> Fixes **incorrect analytic stress** for slabs and other partially periodic systems, where stress was scaled down by the ratio of the artificially inflated vacuum cell to the true cell.
> 
> **`get_neighborhood`** still builds an **extended cell** along non-periodic axes for `matscipy` neighbour binning, but passes that cell only into `neighbour_list`. When **any** direction is periodic, it now **returns a copy of the physical cell** (not the blown-up one) so downstream stress normalization via `det(cell)` uses the real volume. Fully aperiodic systems still return the extended cell, unchanged.
> 
> Tests add a **physical-cell assertion** in `test_half_periodic` and a **`test_stress_partial_pbc`** regression that compares finite-difference strain energy to analytic `sigma_yy` on a Si TTF slab.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1c210e76134fc74faf8f9a6d61577b1259556fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mace/data/neighborhood.py
+++ b/mace/data/neighborhood.py
@@ -20,25 +20,27 @@ def get_neighborhood(
     assert len(pbc) == 3 and all(isinstance(i, (bool, np.bool_)) for i in pbc)
     assert cell.shape == (3, 3)
 
-    pbc_x = pbc[0]
-    pbc_y = pbc[1]
-    pbc_z = pbc[2]
     identity = np.identity(3, dtype=float)
     max_positions = np.max(np.absolute(positions)) + 1
-    # Extend cell in non-periodic directions
-    # For models with more than 5 layers, the multiplicative constant needs to be increased.
-    # temp_cell = np.copy(cell)
-    if not pbc_x:
-        cell[0, :] = max_positions * 5 * cutoff * identity[0, :]
-    if not pbc_y:
-        cell[1, :] = max_positions * 5 * cutoff * identity[1, :]
-    if not pbc_z:
-        cell[2, :] = max_positions * 5 * cutoff * identity[2, :]
+    # blow up the cell in non-periodic directions so matscipy can bin the atoms
+    # (constant may need bumping for models with more than 5 layers)
+    extended_cell = np.array(cell, dtype=float, copy=True)
+    for dim in range(3):
+        if not pbc[dim]:
+            extended_cell[dim, :] = max_positions * 5 * cutoff * identity[dim, :]
+
+    # only use the blown-up cell for the neighbour search, don't return it:
+    # stress divides by det(cell) later, so the fake volume shrinks it (#1509).
+    # keep it for fully aperiodic systems though (no stress there anyway)
+    if any(pbc):
+        cell = np.array(cell, dtype=float, copy=True)
+    else:
+        cell = extended_cell
 
     sender, receiver, unit_shifts = neighbour_list(
         quantities="ijS",
         pbc=pbc,
-        cell=cell,
+        cell=extended_cell,
         positions=positions,
         cutoff=cutoff,
         # self_interaction=True,  # we want edges from atom to itself in different periodic images
--- a/tests/unit/test_data.py
+++ b/tests/unit/test_data.py
@@ -206,9 +206,11 @@ def test_half_periodic():
     atoms = ase.build.fcc111("Al", size=(3, 3, 1), vacuum=0.0)
     assert all(atoms.pbc == (True, True, False))
     config = config_from_atoms(atoms)  # first shell dist is 2.864A
-    edge_index, shifts, _, _ = get_neighborhood(
+    edge_index, shifts, _, cell = get_neighborhood(
         config.positions, cutoff=2.9, pbc=(True, True, False), cell=config.cell
     )
+    # should give back the real cell, not the blown-up one (#1509)
+    assert np.allclose(cell, config.cell)
     sender, receiver = edge_index
     vectors = (
         config.positions[receiver] - config.positions[sender] + shifts
--- a/tests/unit/test_models.py
+++ b/tests/unit/test_models.py
@@ -450,3 +450,76 @@ def test_atomic_virials_stresses():
     assert torch.allclose(
         summed_atomic_stresses, total_stress.squeeze(0), atol=1e-6
     ), f"Sum of atomic stresses (normalized by volume) {summed_atomic_stresses} does not match total stress {total_stress.squeeze(0)}"
+
+
+def test_stress_partial_pbc():
+    """
+    Analytic stress must equal the finite-difference strain derivative of the
+    energy for partially periodic (slab) systems (regression test for #1509).
+    """
+    torch.set_default_dtype(torch.float64)
+    torch.manual_seed(7)
+
+    slab_z_table = tools.AtomicNumberTable([14])
+    model = modules.MACE(
+        r_max=5.0,
+        num_bessel=8,
+        num_polynomial_cutoff=6,
+        max_ell=2,
+        interaction_cls=modules.interaction_classes[
+            "RealAgnosticResidualInteractionBlock"
+        ],
+        interaction_cls_first=modules.interaction_classes[
+            "RealAgnosticResidualInteractionBlock"
+        ],
+        num_interactions=2,
+        num_elements=1,
+        hidden_irreps=o3.Irreps("16x0e + 16x1o"),
+        MLP_irreps=o3.Irreps("16x0e"),
+        gate=torch.nn.functional.silu,
+        atomic_energies=np.array([0.0]),
+        avg_num_neighbors=8,
+        atomic_numbers=slab_z_table.zs,
+        correlation=3,
+        radial_type="bessel",
+    )
+
+    def compute(atoms):
+        atomic_data = data.AtomicData.from_config(
+            data.config_from_atoms(atoms), z_table=slab_z_table, cutoff=5.0
+        )
+        loader = torch_geometric.dataloader.DataLoader(
+            dataset=[atomic_data], batch_size=1, shuffle=False, drop_last=False
+        )
+        batch = next(iter(loader)).to_dict()
+        output = model(batch, training=False, compute_stress=True)
+        return output["energy"].item(), output["stress"].detach().numpy()[0]
+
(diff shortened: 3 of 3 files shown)


aacostadiaz:
@cursor review

aacostadiaz:
@cursor review

ilyes319:
@cursor review

gabor1:
@cursor review

ilyes319:
@cursor review

gabor1:
@cursor review

gabor1:
@cursor review

gabor1:
@cursor review

aacostadiaz:
@cursor review

mace/data/neighborhood.py
```diff
@@ -20,25 +20,27 @@ def get_neighborhood(
...
+    extended_cell = np.array(cell, dtype=float, copy=True)
+    for dim in range(3):
+        if not pbc[dim]:
+            extended_cell[dim, :] = max_positions * 5 * cutoff * identity[dim, :]
+
+    # only use the blown-up cell for the neighbour search, don't return it:
+    # stress divides by det(cell) later, so the fake volume shrinks it (#1509).
+    # keep it for fully aperiodic systems though (no stress there anyway)
+    if any(pbc):
+        cell = np.array(cell, dtype=float, copy=True)
```
aacostadiaz, line 36: Edge case worth considering: a zero row along a non-periodic direction now gives det(cell)=0, so stress comes out as zeros and training with stress in the loss silently NaNs all gradients. Hard to debug when it happens.

```suggestion
        cell = np.array(cell, dtype=float, copy=True)
        # a zero row along a non-periodic direction would make det(cell)=0
        # downstream (stress division, rcell) — keep the extended row there
        for dim in range(3):
            if not pbc[dim] and not cell[dim].any():
                cell[dim] = extended_cell[dim]
```

tests/unit/test_data.py
```diff
@@ -206,9 +206,11 @@ def test_half_periodic():
     atoms = ase.build.fcc111("Al", size=(3, 3, 1), vacuum=0.0)
     assert all(atoms.pbc == (True, True, False))
     config = config_from_atoms(atoms)  # first shell dist is 2.864A
-    edge_index, shifts, _, _ = get_neighborhood(
+    edge_index, shifts, _, cell = get_neighborhood(
         config.positions, cutoff=2.9, pbc=(True, True, False), cell=config.cell
     )
+    # should give back the real cell, not the blown-up one (#1509)
+    assert np.allclose(cell, config.cell)
```
aacostadiaz, line 213: I don't think this assert can catch the bug it references: the old code mutated the passed-in cell in place  and returned that same array, so on the old code this compares the inflated cell with itself and passes. Snapshotting the cell before the call makes it a real regression guard.

```suggestion
    cell_before = config.cell.copy()
    edge_index, shifts, _, cell = get_neighborhood(
        config.positions, cutoff=2.9, pbc=(True, True, False), cell=config.cell
    )
    # should give back the real cell, not the blown-up one (#1509)
    assert np.allclose(cell, cell_before)
```

aacostadiaz (commented):
Thanks for fixing this. I left a few comments but the fix itself looks correct to me.

darthjaja6:
Thanks for incorporating this fix and the regression test into #1533. Since the combined PR supersedes this one and also addresses the review comments here, I'll close this and follow the discussion there.

PR #1529, opened 2026-07-20
Register all parameters in optimizer groups explicitly

steffen-wedig:
I came across an issue in registering module parameters for the optimizer during training of PolarMace models. 

[get_params_options](https://github.com/ACEsuit/mace/blob/ef4743d1d901f16ab0306454b57f211cef2c4d08/mace/tools/scripts_utils.py#L931) registers modules parameters by name, instead of using models.parameters() to get all trainable parameters. Submodules that get_params_options did not know about (`lr_source_maps`, `fukui_source_map`,`field_dependent_charges_maps` etc. in PolarMACE) are silently excluded from the optimizer and stayed at their initialization during training. I verified that this issue is not relevant for FMs, but only for from scratch trained PolarMace models. 

This PR makes the following changes:
- Register optional submodules by name (radial_embedding, pair_repulsion_fn, joint_embedding, embedding_readout, les_readouts, les, lr_source_maps, fukui_source_map, field_dependent_charges_maps, local_electron_energy, layer_feature_mixer), skipping absent or parameter-free ones.
- Raise ValueError naming any trainable parameter no group claims, so future submodules fail loudly at optimizer construction instead of silently not training.

Making these changes surfaced two pre-existing test failures on the develop branch and I updated their fixtures.
They were not raised (as far as I can tell)  because the tests are marked as requiring network connection, but the ci-core.yaml doesnt set allow-network to true, so the CI skips them - not sure if thats intentional. The tests are workflows/test_freeze.py and workflows/test_run_train.py::test_run_train_foundation. Before PR 1244 , when the test fixtures were generated,  the bessel_weights were always reloaded as nn.Parameters (making them accumulate gradients), but they were not added to the optimizer (because no explicit by-name mention in get_params_options) . Gradient clipping however uses model.parameters() irrespective of which parameters are registered in the optimizer, so thereby the tests that test finetuning of foundation models are impacted  no matter if  radial_weights are trained or not.  After PR 1244, the radial_weights are loaded correctly as buffers in mace/tools/finetuning_utils.py::_copy_radial_weights which changes the behaviour of the test, therefore the fixtures had to be regenerated causing small changes in the energies. 

All other training tests pass unchanged, also adding tests/unit/test_optimizer_param_groups.py covering every model class, the trainable-Bessel case, and the unregistered-submodule raise.

Something that should be confirmed is whether zero weight decay is correct for all these modules or if they should have individual lr values, and whether these settings agree with the training setup for the foundation models.  Please lmk if I should change something!


tests/unit/test_optimizer_param_groups.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+
+
+MODEL_BUILDERS = {
+    "MACE": build_mace,
+    "ScaleShiftMACE": build_scale_shift_mace,
+    "AtomicDipolesMACE": build_atomic_dipoles_mace,
+    "AtomicDielectricMACE": build_atomic_dielectric_mace,
+    "EnergyDipolesMACE": build_energy_dipoles_mace,
+    "PolarMACE": pytest.param(
```
aacostadiaz: I think this PolarMACE case never runs on PR CI. The unit job doesn't install `graph_longrange` (so the skipif always skips) and the polar job only collects `tests/extensions/polar`. Could we move it to `tests/extensions/polar/` and use `marks=pytest.mark.polar` instead of the skipif?

mace/tools/scripts_utils.py
```diff
@@ -996,30 +996,39 @@ def get_params_options(
...
-                "weight_decay": 0.0,
-            }
-        )
-    if hasattr(model, "les_readouts") and model.les_readouts is not None:
+    # Optional submodules that only exist on some model classes (joint
+    # embeddings, MACELES, PolarMACE). Each gets its own named group so that
+    # every trainable parameter is registered explicitly; weight decay stays
+    # off because these blocks mix biases, gates, and physically meaningful
+    # scalars that must not be regularized toward zero.
+    optional_submodule_names = [
```
aacostadiaz, line 1004: The optimizer group count goes from 5 to ~10, so `--restart_latest` from a checkpoint made before this change fails to load the optimizer state, and `run_train` swallows that error, so training silently restarts at epoch 0 with a fresh optimizer (verified locally). Worth logging a clear warning when the optimizer state fails.

mace/tools/scripts_utils.py
```diff
@@ -1031,6 +1040,26 @@ def get_params_options(
                 "weight_decay": 0.0,
             }
         )
+
+    # Guard against silently untrained submodules: any trainable parameter
```
aacostadiaz, line 1044: Small caveat: LES builds its Atomwise MLP lazily in the first forward(), after the optimizer is created. So those parameters still slip past the guard. Pre-existing, but maybe worth a sentence in the comment so the guard doesn't read as a complete guarantee.

tests/unit/test_optimizer_param_groups.py
```diff
@@ -0,0 +1,203 @@
...
+        fixedpoint_update_config={
+            "type": "AgnosticEmbeddedOneBodyVariableUpdate",
+            "potential_embedding_cls": "AgnosticChargeBiasedLinearPotentialEmbedding",
+            "nonlinearity_cls": "MLPNonLinearity",
+        },
+        field_readout_config={"type": "OneBodyMLPFieldReadout"},
+    )
+
+
+MODEL_BUILDERS = {
```
aacostadiaz, line 79: MagneticScaleShiftMACE is missing

aacostadiaz (commented):
Thanks for tracking this down. A few comments below.

tests/unit/test_optimizer_param_groups.py
steffen-wedig: Moved both the Polar and Magnetic Mace test to the extension dir.

mace/tools/scripts_utils.py
steffen-wedig: I added error handling to the mace/tools/checkpoint.py::load_checkpoint function, that adds a warning on ValueErrors from optimizer reloading and a regression test that triggers this.

mace/tools/scripts_utils.py
```diff
@@ -1031,6 +1040,31 @@ def get_params_options(
...
+    # that no group claims would never receive optimizer updates.Submodules 
+    # that build their parameters lazily inside the first forward() (e.g. the 
+    # external LES Atomwise MLP (created only when les_arguments
+    # ["use_atomwise"] is True) do not exist yet here, so they slip past both
```
aacostadiaz: Tiny thing, but it blocks pre-commit: these three comment lines have trailing whitespace

```suggestion
    # that no group claims would never receive optimizer updates.Submodules
    # that build their parameters lazily inside the first forward() (e.g. the
    # external LES Atomwise MLP (created only when les_arguments
    # ["use_atomwise"] is True) do not exist yet here, so they slip past both
```

aacostadiaz (approved):
Thanks, this addresses everything. Nice work!

PR #1533, opened 2026-07-22
Fix non-periodic cell in get_neighborhood: extent-based padding + physical cell on partial PBC

aacostadiaz:
## Summary

Combines and reconciles two open fixes to `get_neighborhood`'s handling of the fictitious cell used along non-periodic directions. They touch the same lines and conflict textually, so they are merged into one coherent change. Supersedes #1473 and #1523.

- **#1473**: the `max(abs(positions)) * 5 * cutoff` cell depended on the absolute coordinate origin and produced enormous cells, OOM-ing  PolarMACE's k-space electrostatics for ordinary molecules far from the origin.
- **#1523**: that fictitious cell was *returned* and stored in `AtomicData`, so `stress = virial / det(cell)` was scaled by a large factor  (~18x) for partially periodic (slab) systems.

## What changed

`get_neighborhood` now:

1. Sizes the blown-up cell from the **atom extent** (`extent + 2*cutoff + 1`)  instead of `max(abs(positions)) * 5 * cutoff` — origin-independent, far smaller,  identical neighbour lists (fixes the OOM).
2. Uses that extended cell **only for the matscipy neighbour search**.
3. **Returns the physical cell** whenever any axis is periodic, so stress normalizes by the true volume (fixes the slab stress). Fully aperiodic systems keep the extended cell. The input array is no longer mutated.

## Effect on electrostatic models (PolarMACE)

The returned cell also becomes `AtomicData`'s `volume`/`rcell`, which drive PolarMACE's electrostatics. `graph_longrange` already applies the correct boundary treatment but they **divide by `det(cell)`**. With the old inflated cell that volume was fake, so the corrections were silently scaled away; returning the physical cell gives them the
right volume. So this change is not only a stress fix, it also makes PolarMACE's slab/molecule electrostatic corrections numerically correct. (`graph_longrange` only supports `TTT`/`TTF` and raises otherwise, so no other partial-PBC case is affected.) Adequate vacuum along the non-periodic axis remains the caller's responsibility.

---

> [!NOTE]
> **Medium Risk**
> Touches core graph construction used by all models; slab stress and PolarMACE electrostatics numerics change by design, though behaviour is heavily regression-tested.
> 
> **Overview**
> **`get_neighborhood`** no longer mutates the caller’s cell and fixes how fictitious non-periodic cell rows are built and what gets returned downstream.
> 
> For matscipy neighbour binning only, non-periodic axes use an **extent + cutoff** padded cell (`extent + 2*cutoff + 1`) instead of **`max(|positions|) * 5 * cutoff`**, so padding is origin-independent and much smaller—avoiding huge k-space boxes that could OOM PolarMACE electrostatics while keeping the same neighbour lists.
> 
> When **any axis is periodic**, the function **returns the physical cell** (with a fallback extended row if a non-periodic axis is all zeros), so `AtomicData` volume/`rcell`, slab **stress** (`virial / det(cell)`), and long-range **dipole corrections** use the real box volume. Fully aperiodic systems still return the extended cell.
> 
> Tests cover partial-PBC slabs (PolarMACE cell contract and vacuum convergence), brute-force neighbour parity on skewed slabs, extent-based aperiodic cells, slab stress vs finite differences, and looser cueq stress parity bounds after the smaller cell volume.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6cca0b21114b0bb751b56ea81318bf9e0e431db2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mace/data/neighborhood.py
+++ b/mace/data/neighborhood.py
@@ -88,4 +88,3 @@ def get_neighborhood(
     shifts = np.dot(unit_shifts, cell)  # [n_edges, 3]
 
     return edge_index, shifts, unit_shifts, cell
-
--- a/mace/data/neighborhood.py
+++ b/mace/data/neighborhood.py
@@ -24,16 +24,23 @@ def get_neighborhood(
     pbc_y = pbc[1]
     pbc_z = pbc[2]
     identity = np.identity(3, dtype=float)
-    max_positions = np.max(np.absolute(positions)) + 1
-    # Extend cell in non-periodic directions
-    # For models with more than 5 layers, the multiplicative constant needs to be increased.
-    # temp_cell = np.copy(cell)
+    # Extend cell in non-periodic directions using the actual extent of
+    # the atoms plus cutoff padding.  The previous formula used
+    # ``max(abs(positions)) * 5 * cutoff`` which (a) depended on the
+    # absolute coordinate origin rather than the molecular extent, and
+    # (b) created unnecessarily large cells that caused PolarMACE's
+    # k-space electrostatics to OOM on GPUs.  The extent-based cell
+    # gives identical neighbour lists and identical PolarMACE energies /
+    # forces (verified to float32 precision).
     if not pbc_x:
-        cell[0, :] = max_positions * 5 * cutoff * identity[0, :]
+        extent_x = positions[:, 0].max() - positions[:, 0].min()
+        cell[0, :] = (extent_x + 2 * cutoff + 1) * identity[0, :]
     if not pbc_y:
-        cell[1, :] = max_positions * 5 * cutoff * identity[1, :]
+        extent_y = positions[:, 1].max() - positions[:, 1].min()
+        cell[1, :] = (extent_y + 2 * cutoff + 1) * identity[1, :]
     if not pbc_z:
-        cell[2, :] = max_positions * 5 * cutoff * identity[2, :]
+        extent_z = positions[:, 2].max() - positions[:, 2].min()
+        cell[2, :] = (extent_z + 2 * cutoff + 1) * identity[2, :]
 
     sender, receiver, unit_shifts = neighbour_list(
         quantities="ijS",
@@ -64,3 +71,4 @@ def get_neighborhood(
     shifts = np.dot(unit_shifts, cell)  # [n_edges, 3]
 
     return edge_index, shifts, unit_shifts, cell
+
--- a/mace/data/neighborhood.py
+++ b/mace/data/neighborhood.py
@@ -20,32 +20,49 @@ def get_neighborhood(
     assert len(pbc) == 3 and all(isinstance(i, (bool, np.bool_)) for i in pbc)
     assert cell.shape == (3, 3)
 
-    pbc_x = pbc[0]
-    pbc_y = pbc[1]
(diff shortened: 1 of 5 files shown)


bernstei:
> atom extent (extent + 2*cutoff + 1)

Has anyone carefully thought about whether the implemented solution works when the cell vectors are not orthogonal?  It's easy for naive heuristics to fail under these circumstances

aacostadiaz:
> > atom extent (extent + 2*cutoff + 1)
> 
> Has anyone carefully thought about whether the implemented solution works when the cell vectors are not orthogonal? It's easy for naive heuristics to fail under these circumstances

Good point. I added tests that check the neighbour list on skewed (non-orthogonal) cells: hexagonal, monoclinic and triclinic slabs, with different cutoffs and random atom positions. Each one is compared atom by atom against a brute-force reference and they all match, so the extent-based cell works fine here.

mace/data/neighborhood.py
```diff
@@ -20,25 +20,49 @@ def get_neighborhood(
...
+    # extent (+ cutoff padding) rather than from max(abs(positions)): the old
+    # `max(abs(positions)) * 5 * cutoff` depended on the absolute coordinate
+    # origin and produced huge cells, which blow up PolarMACE's k-space
+    # electrostatics into GPU OOM. Extent-based padding gives identical
+    # neighbour lists at a fraction of the volume.
+    extended_cell = np.array(cell, dtype=float, copy=True)
+    for dim in range(3):
+        if not pbc[dim]:
+            extent = positions[:, dim].max() - positions[:, dim].min()
+            extended_cell[dim, :] = (extent + 2 * cutoff + 1) * identity[dim, :]
```
bernstei, line 36: I want to think about it more, but I don't understand how this can possibly work in general.  It replaces periodic vector a_i with a vector along _Cartesian_ direction i, with 0-x, 1-y, and 2-z, which doesn't necessarily have anything to do with a_i.

bernstei: For example, will this work with
```
[[0, 5, 0],
 [5, 0, 0],
 [0, 0, -5]]
```
and `pbc=[False, True, True]`?  Won't it become
```
[[L, 0, 0],
 [5, 0, 0],
 [0, 0, -5]]
```
where L is the related to the x extent of the system?

If a system is non-periodic along a dimension, I think you need to make that vector along `np.cross(a[i+1 % 3], a[i + 2 % 3])` with some sufficient magnitude, which is not entirely trivial to calculate.  And the possibility of magnitude 0 vectors complicates this a bit more. You probably have to construct 1 or 2 (or 3) vectors that are independent, to use as dummy directions, before doing the cross product.

gabor1:
Can we propose that neighbourlist packages, such as matscipy take on this functionality, of determining the minimum padding?

ilyes319:
I agree that should be the case.

mace/data/neighborhood.py
aacostadiaz: Thanks, you are right. As suggested, this is probably better handled upstream in the neighbour list packages, so I'll close this one for now.

PR #1534, opened 2026-07-22
Allow imports to work if openequivariance is installed but not running on supported GPU

### LINKED ISSUE #1532: openequivariance support breaks cpu torch
Now that I've updated to latest `main` MACE 0.3.16 and installed openequivariance, I can no longer use torch functionality on a CPU.  Initially I discovered it with `symmetrix_extract_mace` (see below), but with CPU torch loaded I can't even import `MACECalculator`
```
>>> from mace.calculators import MACECalculator
/opt/local/python/mace/conda/lib/python3.11/site-packages/e3nn/o3/_wigner.py:10: UserWarning: Environment variable TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD detected, since the`weights_only` argument was not explicitly passed to `torch.load`, forcing weights_only=False.
  _Jd, _W3j_flat, _W3j_indices = torch.load(os.path.join(os.path.dirname(__file__), 'constants.pt'))
Error while loading libcue_ops.so: libcublas.so.12: cannot open shared object file: No such file or directory
  File "<stdin>", line 1, in <module>
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/calculators/__init__.py", line 1, in <module>
    from .foundations_models import mace_anicc, mace_mp, mace_off, mace_omol, mace_polar
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/calculators/foundations_models.py", line 13, in <module>
    from .mace import MACECalculator
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/calculators/mace.py", line 24, in <module>
    from mace.modules.utils import extract_invariant
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/modules/__init__.py", line 5, in <module>
    from .blocks import (
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/modules/blocks.py", line 16, in <module>
    from mace.modules.wrapper_ops import (
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/modules/wrapper_ops.py", line 25, in <module>
    import openequivariance as oeq
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/openequivariance/__init__.py", line 38, in <module>
    from openequivariance._torch.TensorProduct import TensorProduct
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/openequivariance/_torch/TensorProduct.py", line 3, in <module>
    from openequivariance._torch import extlib
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/openequivariance/_torch/extlib/__init__.py", line 23, in <module>
    assert torch.version.cuda or torch.version.hip, (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Only CUDA and HIP backends are supported
```

`symmetrix_extract_mace` used to be able to work on a cpu, with cpu torch, but now it gives an error related to openequivariance.  This seems to be as of a recent update to 0.3.16 and installation of `openequivariance`.  The symmetrix conversion script is unchanged, so I think it'd be perfectly happy to operate without openequivariance.  Is this something that can be fixed at the symmetrix import level, or does there just need to be a better fallback for mace?

```
  File "/opt/local/python/mace/conda/bin/symmetrix_extract_mace", line 5, in <module>
    from symmetrix.cli.extract_mace import main
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/symmetrix/cli/extract_mace.py", line 5, in <module>
    from ..extract_mace_data import extract_mace_data 
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/symmetrix/extract_mace_data.py", line 13, in <module>
    from mace.modules.radial import ZBLBasis
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/modules/__init__.py", line 5, in <module>
    from .blocks import (
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/modules/blocks.py", line 16, in <module>
    from mace.modules.wrapper_ops import (
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/mace/modules/wrapper_ops.py", line 25, in <module>
    import openequivariance as oeq
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/openequivariance/__init__.py", line 38, in <module>
    from openequivariance._torch.TensorProduct import TensorProduct
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/openequivariance/_torch/TensorProduct.py", line 3, in <module>
    from openequivariance._torch import extlib
  File "/opt/local/python/mace/conda/lib/python3.11/site-packages/openequivariance/_torch/extlib/__init__.py", line 23, in <module>
    assert torch.version.cuda or torch.version.hip, (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Only CUDA and HIP backends are supported
```

bernstei:
Check for `AssertionError` that contains a specific string when importing `openequivariance`, and allow it to continue without OEQ if it's caused by running on a non-supported GPU (e.g. torch _CPU_)

closes #1532

### DIFF
--- a/mace/modules/wrapper_ops.py
+++ b/mace/modules/wrapper_ops.py
@@ -27,6 +27,12 @@ try:
     OEQ_AVAILABLE = True
 except ImportError:
     OEQ_AVAILABLE = False
+except AssertionError as exc:
+    if "Only CUDA and HIP backends are supported" in str(exc):
+        # continue w/o if using CPU torch
+        OEQ_AVAILABLE = False
+    else:
+        raise
 
 
 @dataclasses.dataclass
--- a/mace/cli/convert_e3nn_hybrid.py
+++ b/mace/cli/convert_e3nn_hybrid.py
@@ -22,6 +22,12 @@ try:
     OEQ_AVAILABLE = True
 except ImportError:
     OEQ_AVAILABLE = False
+except AssertionError as exc:
+    if "Only CUDA and HIP backends are supported" in str(exc):
+        # continue w/o if using CPU torch
+        OEQ_AVAILABLE = False
+    else:
+        raise
 
 
 def run(


mace/modules/wrapper_ops.py
```diff
@@ -27,6 +27,12 @@
     OEQ_AVAILABLE = True
 except ImportError:
     OEQ_AVAILABLE = False
+except AssertionError as exc:
```
aacostadiaz, line 30: Two test files still have the same crash: `tests/helpers.py` and `tests/backends/backend_parity.py`. In those two files, catching AssertionError too is enough. 

```python
except (ImportError, AssertionError):
    OEQ_AVAILABLE = False
```

aacostadiaz (commented):
Thanks for the fix! One small thing below, otherwise looks good to me.

mace/modules/wrapper_ops.py
bernstei: I'm sure that catching just any generic assertion error _is_ enough, but I was trying to be careful, so that only the particular backend-related assertion was caught, to allow for other failure modes to still lead to an exception.  

Do you not think we should apply that level of specificity to the tests as well? Who knows what other assertions those imports could trigger.

aacostadiaz: That's a fair point and I think you're right. Let's keep your level of specificity in the tests too

PR #1536, opened 2026-07-24
Add control of dispersion coordination cutoff in mace_mp

bernstei:
Add argument to `mace_mp` to allow for control over coordination cutoff (`cnthr` in torch-dftd).


aacostadiaz (approved):
Thanks, this looks good!

Issue #1298, opened 2025-12-02
MH-1 produce exploding results when enabling enable_cueq=True


cpeng77:
**Describe the bug**

For **MACE-MH-1**

`calc = MACECalculator(model_paths = 'mace-mh-1.model', default_dtype = "float64", device = 'cuda', enable_cueq = True, head = 'omat_pbe')`

The energy result created may be exploded, and the optimization of the structure based on this `calc` may not converge.

However, **MACE-MH-1 with enable_cueq = False** produces a reasonable result. 
**MACE OMAT with enable_cueq = True** also produces a reasonable result.
**MACE-MH-1 with device = 'cpu'** also produces a reasonable result.

**To Reproduce**
```python
target_ASE = AseAtomsAdaptor().get_atoms(target_structure)
calc = MACECalculator(model_paths='mace-mh-1.model',default_dtype="float64",device='cuda',enable_cueq=False,head='omat_pbe')
target_ASE.calc = calc
energy = target_ASE.get_potential_energy()
```
This `energy` differs between `enable_cueq = True` and `enable_cueq = False`.

**Results**
H2O without CuEq: -14.227938 eV
H2O with CuEq:    -12.572830 eV
Li3PO4 without CuEq: -200.885284 eV
Li3PO4 with CuEq:    5499.824964 eV

**Expected behavior**
Probably the cuEq is not working well with multi-head MACE?

**Softwares**
cuequivariance                0.7.0
cuequivariance-ops-cu12       0.7.0
cuequivariance-ops-torch-cu12 0.7.0
cuequivariance-torch          0.7.0
mace-torch                    0.3.14

ilyes319:
I need to fix it.


________________________________
From: Peng Cheng ***@***.***>
Sent: Tuesday, December 2, 2025 7:01:50 AM
To: ACEsuit/mace ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [ACEsuit/mace] MH-1 produce exploding results when enabling enable_cueq=True (Issue #1298)

[https://avatars.githubusercontent.com/u/210682797?s=20&v=4]cpeng77 created an issue (ACEsuit/mace#1298)<https://github.com/ACEsuit/mace/issues/1298>

Describe the bug

For MACE-MH-1

calc = MACECalculator(model_paths = 'mace-mh-1.model', default_dtype = "float64", device = 'cuda', enable_cueq = True, head = 'omat_pbe')

[...]

PR #1537, opened 2026-07-26
Test hybrid multi-head backend parity

darthjaja6:
## Summary

- extend the CUDA multi-head parity regression from pure cuEquivariance to the hybrid cuEquivariance/OpenEquivariance conversion
- exercise both DFT and MP2 heads
- compare energy, forces, and stress against the e3nn reference

## Context

Issue #1298 reports large and non-deterministic MACE-MH-1 disagreements with both CuEq and hybrid acceleration. The historical multi-head readout leakage is already fixed, and current tests cover pure CuEq, but the reported hybrid path had no equivalent regression coverage.

## Validation

- `MACE_REQUIRE_CAPS=gpu,cueq,oeq pytest tests/backends -q` — 52 passed
- Black, isort, pylint, and `git diff --check`

### DIFF
--- a/tests/backends/test_cueq.py
+++ b/tests/backends/test_cueq.py
@@ -1,6 +1,7 @@
 """cuEquivariance backend: e3nn parity (CPU or GPU) and the CUDA multi-head
 regression (issue #1298). Capability: cueq; GPU cases carry the gpu marker.
 """
+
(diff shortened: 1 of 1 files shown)


aacostadiaz:
@cursor review

aacostadiaz (approved):

PR #1538, opened 2026-07-26
Checkpoint interaction neighbor normalization

### LINKED ISSUE #1415: Interaction layer 'avg_num_neighbors' is not a Torch buffer.
**Describe the bug**
The interaction layer `avg_num_neighbors` is a regular float, not a Torch buffer. When loading checkpoint weights, it does not overwrite `avg_num_neighbors` of the model, potentially leading to inconsistencies.

**To Reproduce**
Train two identical models, only changing the underlying training dataset. They will have a different `avg_num_neighbors` value, meaning that
```
model1 = torch.load(path_model1)
model2= torch.load(path_model2)

ckpt = torch.load(path_ckpt)

model1.load_state_dict(ckpt["model"])
model2.load_state_dict(ckpt["model"])
```
will not give equivalent results.

This might be happening in every restarted training run that changes the training dataset.

**Expected behavior**
I think it would be more consistent to fully separate model architecture from weights and buffers. This already happens for the `scale_shift` block, which is also data-dependent.

**Additional context**
Not fully related, but it would be nice for a training run to save the initialised model before it starts training. That way, if something happens during the run, you are not left with a bunch of checkpoints and no model to load them into. (see #564)

darthjaja6:
Fixes #1415.

## Summary
- register each interaction block avg_num_neighbors as a persistent buffer, so it follows checkpoints and dtype/device conversions
- keep strict loading compatible with legacy checkpoints that do not contain the new buffer
- update model conversion and fine-tuning transfers to copy legacy float values safely

## Validation
- pytest tests/unit/test_finetuning_utils.py tests/unit/test_models.py -q (15 passed)
- pytest tests/unit -q -k "not benchmark" (247 passed, 16 deselected)
- MACE_REQUIRE_CAPS=gpu,cueq,oeq pytest tests/backends -q (50 passed)
- converted the released legacy MH1 model and verified the float normalization becomes a checkpointed tensor
- black, isort, pylint, and git diff --check

### DIFF
--- a/mace/cli/convert_cueq_e3nn.py
+++ b/mace/cli/convert_cueq_e3nn.py
@@ -220,9 +220,9 @@ def transfer_weights(
 
     # Transfer avg_num_neighbors
     for i in range(num_layers):
-        target_model.interactions[i].avg_num_neighbors = source_model.interactions[
-            i
-        ].avg_num_neighbors
+        target_model.interactions[i].set_avg_num_neighbors(
+            source_model.interactions[i].avg_num_neighbors
+        )
 
     # Load state dict into target model
     target_model.load_state_dict(target_dict)
--- a/mace/cli/convert_e3nn_cueq.py
+++ b/mace/cli/convert_e3nn_cueq.py
@@ -204,9 +204,9 @@ def transfer_weights(
                 )
     # Transfer avg_num_neighbors
     for i in range(num_layers):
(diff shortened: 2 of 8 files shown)


aacostadiaz (approved):
Thanks for this, it looks good.

Your branch had gone conflicting since #1420 landed, so I pushed a merge of develop to it. Both branches added tests at the end of tests/unit/test_finetuning_utils.py, so I kept both sets. The source file merged on its own.

PR #1622, opened 2026-08-05
Fix bug in symmetric contraction

rwkeane:
Fixes an iteration order bug in weighting for the symmetric contraction


aacostadiaz (approved):
Thanks for catching this. I verified the bug on develop before and after: with irreps_in="4x0e + 4x1o" and irrep_out="2e" at correlation=3, U_matrix_1 is all zeros, yet develop zeroes weights[0], which forward() pairs with nu=2, and leaves the dead nu=1 slot trainable. Your mapping is the right one.

Since it changes trained numbers silently, I pushed a regression test to your branch. It fails without your patch.

PR #1630, opened 2026-08-07
Fix Zenodo badge

ElliottKasoar:
Replaces Zenodo badge with custom shields.io badge, as the Zenodo badge has been unreliable for a while. (See https://github.com/scikit-learn/scikit-learn/issues/33840 etc.)


aacostadiaz (approved):
Looks good to me.

PR #1631, opened 2026-08-07
Upload code coverage to coveralls

ElliottKasoar:
This will require someone with more power than me to enable the repo at https://coveralls.io/repos/new (@ilyes319), but from testing, this seems to work for me on my fork: https://coveralls.io/github/ElliottKasoar/mace/?branch=develop, which lines up with the coverage currently that is already generated, but not easily accessible:

image

We may need to change things a little in terms of the core/nightly coverage, as I'm not fully sure what this will look like.


README.md
```diff
@@ -6,6 +6,7 @@
 [![GitHub issues](https://img.shields.io/github/issues/ACEsuit/mace.svg)](https://GitHub.com/ACEsuit/mace/issues/)
 [![Documentation Status](https://readthedocs.org/projects/mace/badge/)](https://mace-docs.readthedocs.io/en/latest/)
 [![DOI](https://zenodo.org/badge/505964914.svg)](https://doi.org/10.5281/zenodo.14103332)
+[![Coverage](https://coveralls.io/repos/github/ACEsuit/mace/badge.svg?branch=develop)](https://coveralls.io/github/ACEsuit/mace/?branch=main)
```
aacostadiaz: Small mismatch here: the badge reads branch=develop but the link goes to branch=main. Since develop is the default branch, I think the link is the one to change.

.github/workflows/ci-core.yaml
```diff
@@ -121,6 +121,14 @@ jobs:
             coverage report --skip-covered --sort=cover | tail -40
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
+      - name: Upload coverage to Coveralls
+        if: always()
```
aacostadiaz: Could we drop the `if: always()` here? pytest-cov still writes coverage.xml when tests fail, so on a failing run this publishes a partial number as the project coverage. That is the same reasoning already written above the nightly's coverage-report job, which is deliberately not always(). Same applies to the copy in `nightly.yaml`

.github/workflows/nightly.yaml
```diff
@@ -245,6 +245,14 @@ jobs:
             coverage report --skip-covered --sort=cover | tail -40
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
+      - name: Upload coverage to Coveralls
+        if: always()
+        uses: coverallsapp/github-action@v2
```
aacostadiaz, line 326: These two uploads measure different scopes: ci-core covers tests/unit + tests/workflows and this one covers the whole suite with every extra installed. On a push to develop both end up reporting against the same commit, so I think the badge would flip between the two numbers depending on which upload landed last. 

Would giving each one a flag-name sort that out, e.g. flag-name: core here and flag-name: nightly-full in ci-core? 

Am I reading the Coveralls behaviour right?

aacostadiaz (commented):
Thanks, having a coverage badge would be great!!

PR #1641, opened 2026-08-15
Fix cuEquivariance fusion device detection

zhubonan:
## Summary
- normalize the conversion device through `torch.device(device).type` before enabling convolution fusion
- preserve existing behavior for string CUDA devices while fixing `torch.device("cuda")`, which is passed by training
- add a focused regression test for both accepted device representations

## Why
The previous string comparison disabled cuEquivariance convolution fusion when the training code passed a `torch.device`. Local RTX 5090 profiling showed the corrected converted route matches native `--only_cueq` throughput within measurement noise.

## Testing
- `python -m pytest tests/unit/test_convert_e3nn_cueq.py tests/backends/test_cueq.py -q` (28 passed)
- `python -m py_compile mace/cli/convert_e3nn_cueq.py tests/unit/test_convert_e3nn_cueq.py`
- `git diff --check origin/develop...HEAD`

### DIFF
--- a/mace/cli/convert_e3nn_cueq.py
+++ b/mace/cli/convert_e3nn_cueq.py
@@ -245,7 +245,7 @@ def run(
         layout=layout,
         group="O3_e3nn",
         optimize_all=True,
-        conv_fusion=(device == "cuda"),
+        conv_fusion=(torch.device(device).type == "cuda"),
     )
 
     # Create new model with cuequivariance config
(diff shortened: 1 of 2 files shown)


zhubonan:
It seems that conv_fusion is detrimental for training as atoms/edges changes every batch. Perhaps it should be turned off by default instead?

aacostadiaz:
@cursor review

aacostadiaz (approved):

PR #1677, opened 2026-08-21
Handle biased e3nn linears when merging LoRA weights

### LINKED ISSUE #1640: Error in LoRA training on MACE-POLAR1
Bug in merging LoRA weights into base model (polar-1-m)

line 88, in _extract_weight_blocks
size = instr.path_shape[0] * instr.path_shape[1]
IndexError: tuple index out of range

MACE version: 0.3.16
CUDA version: 12.6
Torch version: 2.13.0

darthjaja6:
This fixes #1640.

e3nn represents a bias as an instruction with `i_in == -1`, but the actual parameter lives in `linear.bias`, not `linear.weight`. The LoRA merge code was treating that instruction as another weight block, so biased readouts such as the ones in PolarMACE eventually tried to slice past the end of the weight tensor.

This skips bias-only instructions while extracting and composing LoRA weight blocks. The base bias is left unchanged.

I added a regression with a biased `o3.Linear`. The full LoRA unit test file passes (8 tests), and I also loaded the released MACE-POLAR-1-M checkpoint and successfully computed and merged all 44 equivariant LoRA wrappers.

### DIFF
--- a/mace/modules/lora.py
+++ b/mace/modules/lora.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import math
+
 import torch
 import torch.nn.functional as F
 from e3nn import o3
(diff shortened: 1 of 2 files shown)


aacostadiaz (approved):
Thanks for tracking this down, the fix looks right to me

PR #1702, opened 2026-08-27
Disable cuEquivariance convolution fusion for training

zhubonan:
## Summary

- keep convolution fusion enabled by default for normal CUDA conversion and inference;
- add `--cueq_conv_fusion` for training, defaulting to `False`;
- apply the flag to both converted training and native `--only_cueq` model construction.

Users can opt back in with `--cueq_conv_fusion=True` to retest future cuEquivariance releases.

## Motivation

Fusion helps fixed-shape inference, but profiling found a severe regression for variable-shape HDF5 training. cuEquivariance remains enabled; only convolution fusion is disabled by default during training.

## Testing

- focused parser/conversion tests: 27 passed;
- CUDA cuEq backend goldens: 4 passed, 1 skipped (PyTorch 2.10, cuEquivariance 0.9, RTX 5090).

### DIFF
--- a/mace/cli/convert_e3nn_cueq.py
+++ b/mace/cli/convert_e3nn_cueq.py
@@ -220,6 +220,7 @@ def run(
     device="cpu",
     return_model=True,
     layout: str = "ir_mul",
+    conv_fusion: bool = False,
 ):
     # Setup logging
 
@@ -247,7 +248,7 @@ def run(
         layout=layout,
         group="O3_e3nn",
         optimize_all=True,
-        conv_fusion=(torch.device(device).type == "cuda"),
+        conv_fusion=conv_fusion and torch.device(device).type == "cuda",
     )
 
     # Create new model with cuequivariance config
--- a/mace/cli/convert_e3nn_cueq.py
+++ b/mace/cli/convert_e3nn_cueq.py
@@ -220,7 +220,7 @@ def run(
     device="cpu",
     return_model=True,
     layout: str = "ir_mul",
-    conv_fusion: bool = False,
+    conv_fusion: bool = True,
 ):
     # Setup logging
 
--- a/mace/cli/run_train.py
+++ b/mace/cli/run_train.py
@@ -888,7 +888,9 @@ def run(args) -> None:
             "MagneticScaleShiftMACE",
             "AtomicDielectricMACE",
         ]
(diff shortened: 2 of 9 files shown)


mace/cli/convert_e3nn_cueq.py
```diff
@@ -220,6 +220,7 @@ def run(
     device="cpu",
     return_model=True,
     layout: str = "ir_mul",
+    conv_fusion: bool = False,
```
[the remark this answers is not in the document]
zhubonan: Verified: both golden routes failed exactly as described. I changed the policy so normal CUDA conversion remains fused by default (`conv_fusion=True`), while `run_train` explicitly passes `False`. This restores the direct-converter and `MACECalculator(enable_cueq=True)` golden contracts without calculator-specific plumbing. The two affected GPU goldens now pass, and the focused policy/golden suite reports 12 passed, 3 skipped.

mace/cli/run_train.py
```diff
@@ -888,7 +888,9 @@ def run(args) -> None:
             "MagneticScaleShiftMACE",
             "AtomicDielectricMACE",
         ]
-        model = run_e3nn_to_cueq(deepcopy(model), device=device)
+        model = run_e3nn_to_cueq(
+            deepcopy(model), device=device, conv_fusion=False
```
aacostadiaz: Could this be a flag instead of hardcoded? Useful for anyone who wants to re-test fusion later without editing the source.

Also, does `--only_cueq` need the same fix? That path builds the model in `configure_model`, where `model_script_utils.py:187` still switches fusion on for cuda so it would still train fused.

aacostadiaz (commented):
Thanks for this! A couple of questions below.

mace/cli/run_train.py
zhubonan: Implemented in 082804d. Training now exposes `--cueq_conv_fusion` (default `False`) and forwards it through both paths: converted e3nn→cuEq training and native `--only_cueq` construction. The native route also guards the opt-in with `torch.device(args.device).type == "cuda"`. Normal conversion/calculator/TorchSim inference remains fused by default. Focused tests pass (27 unit tests; CUDA cuEq goldens 4 passed, 1 skipped).

aacostadiaz (approved):
Thanks!!

PR #1711, opened 2026-09-05
Avoid the [B, mul1, mul2] intermediate in SparseUvuTensorProduct (up to 2.6x less memory)

lollcat:
`SparseUvuTensorProduct`'s `l x l -> 0` path built a `[B, mul1, mul2]` outer product and
contracted it away on the very next line. Contracting the weight into `x2` first computes the
same thing with an intermediate no larger than the inputs.

Every `SparseUvuTensorProduct` in the MACE-Polar checkpoints is 512 channels wide
(`512x0e ⊗ 512x0e -> 512x0e` in `polar-1-s`), so `[B, 512, 512]` is the peak allocation of a
PolarMACE forward — a full 512×512 outer product per atom, summed away immediately.

### What it buys you

This OOMs on a 16 GB GPU before the change and runs after it:

```python
from mace.calculators import mace_polar

water = molecule("H2O")
water.set_cell([4.0, 4.0, 4.0])
water.pbc = True
atoms = water * (10, 10, 10)  # 3000 atoms
atoms.info["charge"] = 0
atoms.info["spin"] = 1

atoms.calc = mace_polar(
    model="polar-1-s", device="cuda", default_dtype="float64", enable_cueq=True
)
print(atoms.get_potential_energy())
```

Largest box that evaluates energy + forces, bisected on a Tesla T4 (16 GB), float64, that
same periodic water system:

| | before | after | peak MiB/atom |
|---|---:|---:|---|
| `polar-1-s`, stock install | 1368 atoms | **1887** | 9.21 → 8.32 |
| `polar-1-s`, `enable_cueq=True` | 1914 atoms | **4380** | 8.00 → **3.11** |
| `polar-1-m`, `enable_cueq=True` | 1593 atoms | **3393** | 9.43 → 4.07 |
| `polar-1-l`, `enable_cueq=True` | 1224 atoms | **1917** | 12.26 → 7.43 |

cueq gains most because it fuses the interaction blocks, which leaves this un-accelerated
block as the peak allocation; without it the peak sits elsewhere and only the transient outer
product disappears. Drop `enable_cueq=True` from the script and `(8, 8, 8)` — 1536 atoms — is
the equivalent before/after size on a stock install. Energy + forces is also 2.1–2.5× faster
across the three checkpoints with cueq.

### Correctness

The two orders are the same contraction reassociated, and `opt_einsum` picks this one for the
equivalent `uv,zui,zvi->zu` form. Verified against `e3nn.o3.TensorProduct`:

- values and gradients agree to roundoff across float32/float64, both `mul_ir` and `ir_mul`
  layouts, both supported path types, multiplicity 1–256 and `l` up to 3;
- `gradcheck` and `gradgradcheck` pass, and weight gradients and double backward — the
  fine-tuning and force-training paths — match to the same roundoff;
- end to end on `polar-1-s/m/l`: `dE = 0` and `max|dF| <= 2.4e-15` in float64. Every energy
  in the table above is identical before and after at each size that fits both.

### Tests

`SparseUvuTensorProduct` had no tests. This adds them against the `e3nn.o3.TensorProduct` it
stands in for — both path types, both layouts, values and gradients (inputs *and* weights) —
plus a guard that fails if the contraction order regresses, which is otherwise silent until it
OOMs. They import nothing but `torch` and `e3nn`, so they sit in `tests/unit` and run without
the polar extra.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mace/modules/field_blocks.py
+++ b/mace/modules/field_blocks.py
@@ -445,11 +445,16 @@ class SparseUvuTensorProduct(torch.nn.Module):
             # mode_code determines whether out is scalar (d_out=1) or vector (d_out=d1).
             if mode_code == 0:
                 # (l x l -> 0): [B, mul1, d] · [B, mul2, d] -> [B, mul1]
+                # Contract the weight into x2 before contracting over d. Summing over v last
+                # would build a [B, mul1, mul2] intermediate only to contract it away on the
+                # next line; with multiplicities in the hundreds and d = 2l+1 in {1,3,5} that
+                # dominates the memory of the whole Polar model. Same leading flop count, and the same
+                # order e3nn's opt_einsum pass picks for its `zuv,zui,zvi->zu` form.
                 x1v = to_mul_ir(in1_block, mul1, d1)
                 x2v = to_mul_ir(in2_block, mul2, d1)
                 w = self.weight[w_start:w_stop].view(mul1, mul2)
-                pair = torch.einsum("bud,bvd->buv", x1v, x2v) / math.sqrt(float(d1))
-                mixed = torch.einsum("buv,uv->bu", pair, w)
+                weighted_x2 = torch.einsum("uv,bvd->bud", w, x2v)
+                mixed = (x1v * weighted_x2).sum(dim=-1) / math.sqrt(float(d1))
                 out[:, out_start:out_stop] = out_block + path_weight * mixed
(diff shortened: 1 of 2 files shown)


aacostadiaz (approved):
Thanks!!

Issue #1636, opened 2026-08-10
FM-4 — Artifact resolution, download and cache contract


aacostadiaz:
**Depends on:** INF-1 (#1550) · **Blocks:** FM-3 (#1588)

**Prerequisite:** the Phase 0 characterization suite is committed — goldens, fixtures and the single-source tolerance table under `tests/golden/`, characterization tests under `tests/unit/`. References of the form `P0-N` name that work.

**Context:** Getting the bytes of a foundation model onto disk is a tested behavioural surface in its own right, and today it is duplicated five times inside `mace/calculators/foundations_models.py` with real divergence between the copies. `_urlretrieve_with_timeout` (`:41-84`) is the shared part: a 120 s per-read socket timeout (`_DOWNLOAD_TIMEOUT`, `:16`), streaming to a `.part` sibling and `os.replace`-ing on success so an interrupted download never leaves a truncated file that later fails inside `torch.load` with a confusing zip-archive error, a progress line, and removal of the partial on failure. `_normalize_github_download_url` (`:19-38`) rewrites `github.com/{org}/{repo}/{blob|raw}/{ref}/{path}` to `raw.githubusercontent.com` and passes everything else through — including the release-asset URLs (`…/releases/download/…`), whose third path segment is `releases`, so today it affects exactly one URL in the file. Everything above that line diverges:

- **Two different cache-filename schemes.** `download_mace_mp_checkpoint` (`:169-172`) and `download_mace_polar_checkpoint` (`:211-214`) strip every character that is not alphanumeric or `_`, producing `20231210mace128L0_energy_epoch249model` and `MACEPOLAR1Smodel`. `mace_off` (`:411`), `mace_omol` (`:561`) and `mace_mdp` (`:645`) take `basename(url).split("?")[0]`, producing `MACE-OFF23_medium.model` and `MACE-MDP.model`. Both schemes are live in every existing `~/.cache/mace`.
- **The HTML-poison guard is not universal.** `mace_mp`, `mace_polar` and `mace_mdp` check the response for `Content-Type: text/html` and raise; `mace_off` and `mace_omol` do not, so a login page or an error page is cached as a model file.
[...]

PR #1712, opened 2026-09-05
[codex] Reject HTML before caching model downloads

LOGO127:
## Problem

An HTTP 200 response can still be an HTML login/error page. The common downloader currently saves such a response as a model file. Some loaders have a post-download string guard, but it is case-sensitive; OFF, OMOL and ANI-CC do not share that guard. Once an invalid file is cached, subsequent calls may reuse it instead of retrying the download.

## Cause and fix

Move response-header inspection into `_urlretrieve_with_timeout`, before reading the response body or opening the temporary file. Normalize the Content-Type media type (case and optional parameters), and raise the existing style of download error for `text/html`. Binary responses, including those without Content-Type, retain the existing timeout, progress, atomic replacement and return-value behavior.

This is a small v0.3 bug fix against `develop`, following the current contribution guide. It relates to download robustness noted in #1636, but does **not** implement or close the v1 FM-4 artifact/cache redesign. Existing loader-specific guards remain unchanged. The fix does not inspect arbitrary model bytes or repair previously cached invalid files.

## Regression coverage

Fifteen new offline cases exercise the real downloader and all six public loaders (MP, Polar, OFF, OMOL, MDP and ANI-CC). They cover canonical/parameterized/mixed-case HTML media types, case-insensitive header lookup, no body read on rejection, preservation of an existing destination, retry after an HTML response, reuse after a successful download, exact binary bytes, missing Content-Type, timeout/progress/header behavior and interrupted-download cleanup.

HTTP responses are fakes and calculator construction is mocked in the new loader tests; no remote model weights or GPU are required. Existing adjacent calculator/configuration tests are also included.

## Validation

On original production code, the new tests produced **12 failures and 3 passes** (missing HTML rejection). After the fix:

```bash
python -m pytest tests/unit/test_download_http.py tests/unit/test_download_cache.py tests/unit/test_download_urls.py tests/unit/test_calculator_mdp.py tests/unit/test_foundations_models.py -q --tb=short
# 37 passed, 226 warnings in 5.62s

pre-commit run --files mace/calculators/foundations_models.py tests/unit/test_download_http.py
black --check mace/calculators/foundations_models.py tests/unit/test_download_http.py
isort --check-only mace/calculators/foundations_models.py tests/unit/test_download_http.py
```

All applicable selected-file hooks pass. Black/isort were additionally run explicitly on the new test file because most repository hooks exclude tests. No mypy result is claimed. Environment: Ubuntu/WSL, Python 3.12.3, torch 2.11.0+cpu, numpy 2.3.5, e3nn 0.4.4. Warnings include dependency deprecations and the existing host's TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD setting; this patch changes neither that setting nor checkpoint loading policy.

No full test-suite, GPU, real remote checkpoint, training, accuracy or performance result is claimed. No golden references or tolerances changed.

Implementation, regression tests and this description were developed with AI assistance.

### DIFF
--- a/mace/calculators/foundations_models.py
+++ b/mace/calculators/foundations_models.py
@@ -64,10 +64,13 @@ def _urlretrieve_with_timeout(url, filename, timeout=_DOWNLOAD_TIMEOUT):
         with urllib.request.urlopen(
             _normalize_github_download_url(url), timeout=timeout
         ) as response:
+            info = response.info()
+            content_type = info.get("Content-Type", "").split(";", 1)[0].strip().lower()
+            if content_type == "text/html":
+                raise RuntimeError(f"Model download failed, please check the URL {url}")
             total = int(response.headers.get("Content-Length", 0))
             downloaded = 0
             block_size = 256 * 1024  # 256 KB
-            info = response.info()
             with open(tmp, "wb") as out:
                 while True:
                     block = response.read(block_size)
--- a/tests/unit/test_download_http.py
+++ b/tests/unit/test_download_http.py
@@ -1,0 +1,143 @@
+"""Exercise the real checkpoint downloader with offline HTTP responses."""
+
+import io
+from email.message import Message
+from pathlib import Path
+from unittest.mock import Mock
+
+import pytest
+
+from mace.calculators import foundations_models as fm
+
+
+class Response(io.BytesIO):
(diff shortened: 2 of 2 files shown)


aacostadiaz (approved):
Thanks for fixing this!!

PR #1718, opened 2026-09-08
update to graph_longrange v0.4.3

WillBaldwin0:
This changes the interface to `graph_longrange` to use v0.4.3 and fixes a bug in the mace polar stress calculation #1642 .

Important changes:
- The new graph electrostatics interface requires that users specify the electrostatic boundary condition handling explicitly. This means the calculator is now initialised with an argument `pbc_handling`. 
- The newer graph electrostatics classes also have some added attributes. These are patched in when loading a `PolarMACE` model via `__setstate__`. 
- The stress is fixed, and some new, tighter tests were added.
- Requirements updated, I can make another tiny PR for the docs branch.

@aacostadiaz please let me know what you think.


aacostadiaz (approved):
Thanks for this. I pulled this branch and checked it against the graph_electrostatics docs. The fix looks right. Just one comment.

The test only covers one setting, the fixture uses `pbc_handling="pbc"`, but `slab`, `molecule_in_box` and `mixed_periodic` each do extra work that could have the same bug. I ran your finite-difference check on them too and they all pass with room to spare, so it looks like the fix does cover them. Would it be worth adding tests for them?

WillBaldwin0:
Do you mean the stress finite difference test?

aacostadiaz:
> Do you mean the stress finite difference test?

Yes, `test_polar_stress_matches_finite_difference`. `molecule_in_box` cannot be checked this way, so only `slab` and `mixed_periodic` would be worth adding.

WillBaldwin0:
I think the stress is not really well defined for a slab either, since if you are asking for stress you probably want the energy change corresponding to a 2d, in-plane, expansion of the cell, which has different units. I think this is related approximately to the full stress we currently compute, but not exactly because of the dipole correction not being exact for finite cells. 

I would lean towards adding the tests for all the modes just so that we catch any issues with derivatives, and maybe add a comment or something in the docs about stresses in non-pbc cells. What do you think?

PR #1723, opened 2026-09-11
Add typed model outputs and a declarative observable specification

### LINKED ISSUE #1555: CORE-1 — Typed outputs + declarative observable specification
**Depends on:** INF-1 (#1550) · **Blocks:** ARCH-3 (#1562), TRN-2 (#1576), CFG-1 (#1574)

**Context:** Today model outputs are unvalidated dicts. **Eleven** legacy model classes return `Dict[str, Optional[torch.Tensor]]` from `forward` — `MACE`, `ScaleShiftMACE`, `AtomicDipolesMACE`, `AtomicDielectricMACE`, `EnergyDipolesMACE` (`mace/modules/models.py:289,471,778,1065,1345`) and `MACELES`, `PolarMACE`, `MagneticMACE`, `MagneticScaleShiftMACE`, `MagneticSCFMACE`, `TimeReversalSymmetrizedMACE` (`mace/modules/extensions.py:291,966,1707,1759,2029,2278`) — the last of which is not CLI-reachable and is the one a hand-written enumeration drops — and between them they emit **43 distinct string keys**, from `energy`/`forces`/`stress` through `magforces`, `BEC`, `latent_{charges,dipoles,alphas,kappas,quads}`, `external_field`, `fukui_functions`, `polarizability_sh` and the SCF trio `scf_energy_history`/`scf_steps`/`equilibrated_magmom`. Only **40** of the 43 are keys of a returned dict literal: the SCF trio is assigned onto the wrapped model's output after the fact (`extensions.py:2107-2112`), so a count read off return literals stops at 40 and silently loses the SCF surface. And the model forward is only the innermost of **three** layers users can see — the calculator and the eval CLI each add names of their own (constraint 7). Every consumer keeps its own key lists by hand: the ASE calculator maintains hardcoded `graph_level_keys` / `atom_level_keys` sets to slice padding (`mace/calculators/mace.py:463-491`, applied at `:496-499`) which classify **22 of those 43 keys** — the other 21 fall through the `else` branch at `:500-501` and are returned with their padding rows intact. Which observables a model computes is decided by a string ladder: `mace/cli/run_train.py:588-625` branches on the model-class name (`AtomicDipolesMACE`, `AtomicDielectricMACE`, `EnergyDipolesMACE`, `PolarMACE` + loss name) and sets six `compute_*` booleans per branch, while the forward signatures carry ten more (`compute_force`, `compute_virials`, `compute_stress`, `compute_displacement`, `compute_hessian`, `compute_edge_forces`, `compute_atomic_stresses`, `compute_magforces`, `compute_bec`, `compute_dielectric_derivatives`, plus `training` and `lammps_mliap`) — so adding an output property means editing multiple files across the codebase. This ticket builds the two abstractions that replace all of that — a typed, framework-generic output structure and a declarative observable spec. They are the reforge's most consequential new abstractions and feed the whole downstream chain: spec → automatic head creation (ARCH-3 (#1562)) → automatic loss term (TRN-2 (#1576)) → training config (CFG-1 (#1574)). Keep them minimal — no abstraction "for the future".

**Interface & constraints:**

1. **`MACEOutput(Generic[TensorT])`** in `mace_core.outputs`. `TensorT = TypeVar(...)` so the same class carries `torch.Tensor` in mace-torch and `jax.Array` in mace-jax — `mace_core` itself imports neither (a minimal `Protocol` for the tensor type is allowed if needed). Optional core fields: `total_energy`, `node_energies`, `forces`, `stress`, `virials`, `dipole`. Everything else goes through the escape hatch `extras: dict[str, TensorT]` — custom data never breaks the strict typing of core fields. The 43-key legacy surface is what makes `extras` load-bearing rather than decorative: six core fields plus a typed hatch is the deliberate split, and the electrostatic, magnetic and dielectric families live in the hatch with their `per_atom`/`per_graph` classification carried by their `ObservableSpec`, not by a second hardcoded key set. Models never return raw dicts. Implement as a plain dataclass unless validation is actually needed — justify the dataclass-vs-Pydantic choice in the PR. This `Generic[TensorT]` trick is load-bearing beyond this ticket: the kernel Protocol reuses exactly the same pattern to stay framework-free, so it must genuinely work with no framework installed.

2. **`ObservableSpec`** in `mace_core.observables`, a Pydantic model with fields: `name`, `irreps` (string spec of the spherical-tensor shape in the canonical irreps grammar — e.g. `"1o"` for a per-atom vector, `"128x0e+128x1o+128x2e"` for a sum of multiplicities; this ticket stores and grammar-validates the string, it does not implement irreps algebra), `per_atom` (bool), `units` (project convention: eV, Å), `normalization`, `default_loss_weight` (seeds the automatic loss term in TRN-2 (#1576)). **`normalization`** is a required string enum — one of `"none"`, `"std"`, `"rms"` — naming how the target quantity is scaled before it reaches a head and a loss term; it is the per-observable successor of the legacy scaling registry (`std_scaling`, `rms_forces_scaling`, `rms_dipoles_scaling`, the three entries of `scaling_classes` at `mace/modules/__init__.py:91-95`). This ticket only stores and validates the value: the head applies it in the output layer and the loss applies the matching term, both reading this one field, so no second scaling mechanism appears anywhere. Any atomic or total spherical-tensor property declared this way must become trainable with zero new code downstream.

3. **Derivative auto-naming, over declared inputs — not just position and cell.** For a declared quantity `<q>` differentiated with respect to a declared input `<x>`, the derivative is auto-named `d_<q>_d_<x>`, with `pos` and `cell` as the two inputs every model has. Three named special cases, all data rather than code branches scattered in consumers:

| quantity | input | canonical name | sign |
|---|---|---|---|
| `energy` | positions | `forces` | −1 |
| `energy` | cell (strain) | `stress` | +1 |
| `energy` | `magmom` | `magforces` | −1 |

   The third is not an extrapolation: develop computes `magforces` as `−∂E/∂magmom` in the same autograd call as forces (`mace/modules/utils.py:262-299`, `compute_forces_magforces`, returning `-forces, -mag_forces`), it is a first-class loss term (`mace/modules/loss.py:453-458,483-488`) and the SCF model drives its fixed point with it (`extensions.py:2077`). A grammar that only knows `d_<q>_d_pos` and `d_<q>_d_cell` cannot express the magnetic family at all, and MAG-1 (#1584) would then have to bypass the spec — which is the failure this abstraction exists to prevent. Declaring an input feature (`magmom` today; `elec_temp`, `total_charge`, `total_spin` and the `embedding_specs` features tomorrow) and asking for its derivative must require zero new code.

4. **Defaults file.** Ship `defaults/observables.yaml` declaring energy plus its position and cell derivatives — the canonical example every other observable copies — with a loader and tests.

5. **Error contract, and the padding classification.** Validation errors must name the offending observable and state the accepted grammar (agent-friendly: name the key/value and the fix). The downstream hard error — declaring a property that is absent from the data — is enforced at head creation (ARCH-3 (#1562)); this ticket's spec must carry enough metadata (`name`, `per_atom`, `irreps`) to make that check and the automatic padding/unpadding of graph-level vs node-level fields possible. That consumer is DATA-3 (#1571)/DEP-1 (#1583), and the bar it has to clear is explicit: the derived classification must cover **every** key the model emits, so the 21-key fall-through in `mace/calculators/mace.py:500-501` cannot recur. Unclassifiable output is an error at spec time, never a silent pass-through of padded rows.

6. **Purity is tested, not assumed.** Tests exercise `MACEOutput` with numpy arrays (proves framework-agnosticism), and one test asserts `"torch" not in sys.modules` after importing `mace_core`.

7. **The user-observable output surface is three layers, and the completeness gate is defined over all three.** "v1 has the same functionality as develop" is a claim about the names a user can read out, not about the names one function happens to return — a key that reaches the user only through the calculator or only through `mace_eval_configs` is functionality just the same, and a gate keyed on the model forward alone cannot see it. The set is therefore defined as the union of:

   - **(a) model forward** — every string key of the dict returned by `forward` on the eleven model classes in `mace/modules/{models,extensions}.py`, counting both dict-literal keys and keys assigned onto the returned object before it is returned. **43 keys.** Nine of the eleven contribute: `MagneticMACE.forward` (`extensions.py:1697-1708`) is abstract and raises `NotImplementedError`, and `TimeReversalSymmetrizedMACE.forward` (`extensions.py:2278`) returns the wrapped model's keys unchanged, so neither adds a name.
   - **(b) ASE calculator `results`** — every key written into `self.results` by `MACECalculator.calculate` (`mace/calculators/mace.py:727-890`) and `MagneticMACECalculator.calculate` (`:1444-1450`). **31 keys**, of which **15 exist at this layer only**: `free_energy` and `energies` (the aliases and the E0-inclusive per-atom energy, `:801-804`), `stresses` (the Voigt per-atom rename of the model's `atomic_stresses`, `:735-739`), `LES_alphas` / `LES_kappas` (`:832-839`), `bec` (the lower-cased mean of the model's `BEC`, `:840`), `MACE_magmoms` (`:1467`), and the eight committee keys `{energy,forces,stress,dipole}_{comm,var}` (`:771,:781`). Read the committee set off the *code*, not off `implemented_properties`: the loop emits both suffixes for all four members of `results_store_ensemble`, while `implemented_properties` advertises only four of the eight (`:278-293`) — so `forces_var`, `stress_comm` and `dipole_comm` are produced and never declared.
   - **(c) `mace_eval_configs` output** — every name written into `atoms.arrays` or `atoms.info`, after the `--info_prefix` prefix (default `"MACE_"`, `mace/cli/eval_configs.py:103-107`), at `eval_configs.py:423-472`. **13 names** over 14 write sites (`descriptors` is written to `info` or to `arrays` depending on `--descriptor_aggregation_method`), of which **3 exist at this layer only** and each is a *rename* of a model key, which is exactly why they are easy to lose: `BO_contributions` (model `contributions`), `descriptors` (model `node_feats`, after invariant extraction and layer truncation), `node_energies` (model `node_energy`). Only `energy`, `forces` and `stress` also appear as calculator `results` keys — the other **ten reach the user through this CLI alone**: the three renames plus `BEC`, `latent_charges`, `latent_dipoles`, `latent_kappas`, `latent_alphas`, `latent_quads` and `magforces`, for which the calculator offers at most a differently-named subset (`bec`, `LES_alphas`, `LES_kappas`) and usually nothing.

   **Union: 61 distinct names** — 43 + 15 + 3. That number, not 43, is what "the output surface survived the rewrite" is measured against.

   **Ownership, one ticket per layer.** This ticket owns (a): the `ObservableSpec` completeness test enumerates the 43 mechanically and maps each to a spec or a DROP row. **DEP-1 (#1583)** owns the energy-family part of (b) and **DEP-1a (#1634)** the dipole/dielectric/polar/LES/magnetic part; each must account for its layer's calculator-only names against the spec rather than re-growing a hand-kept list. **CLI-1 (#1579)** owns (c) for `mace eval`, with the P0-5 eval contract as its behavioural floor. The three enumerations are the same mechanical extraction run at three sites, so each ticket re-derives its layer from the source rather than trusting a number copied from here.

**Task:**
1. **`mace_core.outputs`:** implement `MACEOutput(Generic[TensorT])` per constraint 1 (dataclass unless validation is needed — justify the choice in the PR); no torch/jax imports; numpy-based tests plus the `sys.modules` purity assertion.
2. **`mace_core.observables`:** implement the Pydantic `ObservableSpec` per constraints 2–3, including the `d_<q>_d_<x>` derivative auto-naming with the three named special cases and their signs; ship `defaults/observables.yaml` declaring energy + position/cell derivatives, with loader and tests; validation errors name the observable and the accepted grammar.
3. Add the completeness test: the 43 legacy output keys enumerated from `mace/modules/{models,extensions}.py` — enumerated mechanically, covering keys assigned onto the output as well as dict literals, so the SCF trio is in the set — each map to exactly one `ObservableSpec` shape (`per_atom` true/false) or to an explicit DROP row with a reason; no key is unaccounted for. Ship the enumeration as a reusable helper, since DEP-1 (#1583)/DEP-1a (#1634) and CLI-1 (#1579) run the same extraction over layers (b) and (c) of constraint 7. Keys DROPped here rather than specified: the SCF diagnostics `scf_energy_history` / `scf_steps` (run telemetry, not observables — they belong in the SCF stage's log, and MAG-1 (#1584) carries that decision), and `equilibrated_magmom`, which is an *input* the SCF stage converged, exposed through the stage result rather than as a model observable.

**Out of scope:** head creation (ARCH-3 (#1562)), loss construction (TRN-2 (#1576)), config machinery (CORE-2 (#1556)), the SCF fixed-point stage itself (MAG-1 (#1584)/TRN-2 (#1576)).

**Acceptance criteria:**
- [ ] Declaring a new rank-2 per-atom observable purely in YAML parses to a valid spec with derived `d_<q>_d_pos` / `d_<q>_d_cell` names, with zero code changes.
- [ ] Declaring a new *input* feature and asking for `d_energy_d_<feature>` parses with zero code changes; the `magmom` → `magforces` (−1) case is covered by a test alongside forces and stress.
- [ ] The forces/stress special case (naming and the −1 sign on the position derivative of energy) is covered by tests.
- [ ] All 43 legacy model-forward output keys are either specified or carry an explicit DROP row with a reason (completeness test), the 43 re-derived from the source by the test rather than hard-coded, so a new key added to a legacy forward fails it.
- [ ] The three-layer surface of constraint 7 is written down where the downstream tickets read it: the 61-name union, its 43/15/3 split, and which ticket owns each layer.
- [ ] `mace-core` imports without torch installed; a test asserts `"torch" not in sys.modules` after importing it; `MACEOutput` tests run on numpy arrays.
- [ ] A malformed spec produces a validation error that names the observable and the accepted grammar.

**Verify:**
```bash
python -m pytest packages/mace-core/tests/test_outputs.py packages/mace-core/tests/test_observables.py -v
ruff check packages/mace-core && ty check packages/mace-core
```
**Review focus:** cross-review — each half (outputs, observables) reviewed by someone who didn't write it. Watch for scope creep: these two abstractions must stay minimal, and the derivative grammar generalising past `pos`/`cell` is the one place where the extra generality is paid for by an existing model family, not by speculation.

---

aacostadiaz:
`MACEOutput` replaces the dict every legacy model forward returns: six core fields plus a typed `extras` hatch, generic over the tensor type so `mace_core` imports no framework. A plain dataclass, since the values are tensors whose type this package cannot name and there is nothing to validate.

`ObservableSpec` makes a property a row instead of a code change. Derivatives are named `d_<q>_d_<x>`, with `forces`, `stress` and `magforces` keeping their own names. The grammar runs over declared inputs, not just positions and the cell, because `magforces` already needs the third. `defaults/observables.yaml` ships energy with its two derivatives.

All 43 keys the frozen forwards emit are classified in `tests/architecture/observable_coverage.py`, read out of the source by the P0-5 scan rather than listed, so a new legacy key fails the test. 22 carry a literal irreps string, 6 state it with the model parameter that sets the rest (readout count, maximum multipole order, whether an anisotropic readout is declared), 5 are derivatives, 10 are drops naming what owns them instead. Nothing is deferred: a row that cannot say what its irreps are fails the test rather than acquiring a TODO.

`docs/reforge/output_surface.md` records the three-layer surface: 43 + 15 + 3 = 61 names, every number re-derived by the test.

Three things the sign checks turned up, all measured on the committed anchors:

- `virials = -dE/dstrain` but `stress = +dE/dstrain / V`. They differ by a sign,  not only the volume (`max|stress*V + virials|` = 1.2e-35).
- `hessian` is `+d2E/dpos2`, the negative of `d(forces)/d(pos)`. It is a second derivative, so it gets a Drop row naming the derivative engine.
- `BEC` is not another spelling of `dmu_dr`. LES differentiates a Berry-phase polarization built from mean-removed latent charges; the dielectric model differentiates its `dipole` readout. Different quantity, different unit, measured shape `(n_atoms, 2, 3, 3)`.

Implements CORE-1 (#1555).

---

> [!NOTE]
> **Medium Risk**
> Adds a foundational contract layer with strict validation on derivative signs/names and completeness tests tied to legacy output keys; mistakes here would propagate to training, calculators, and parity gates, but runtime legacy paths are unchanged until backends adopt these types.
> 
> **Overview**
> Introduces the **mace-core** contract for v1 model outputs and observables: **`MACEOutput`** replaces the legacy forward dict with six core fields plus typed **`extras`**, generic over tensor type so the package stays framework-free.
> 
> A **declarative observable catalogue** loads from YAML (`defaults/observables.yaml` for energy, forces, stress): Pydantic **`InputSpec` / `ObservableSpec` / `ObservableCatalogue`**, irreps grammar parsing, and derivative naming (`d_<q>_d_<x>` with explicit names/signs for forces, stress, magforces). **`pyproject.toml`** now depends on numpy, pydantic, and PyYAML.
> 
> **Architecture gates** classify all **43** legacy model forward keys in `observable_coverage.py` and assert completeness via `surface_scan` plus **`test_observable_completeness`** (including the **61-name** union across model, ASE calculator, and eval CLI documented in **`output_surface.md`**). Reforge docs (`target_layout.md`, `extending_mace.md`) are updated to describe table-row observables and **`LossConfig`** weights instead of `default_loss_weight` in YAML.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86080f5e1e425e3247113be75c2520072fb64b97. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/mace-core/src/mace_core/defaults/observables.yaml
+++ b/packages/mace-core/src/mace_core/defaults/observables.yaml
@@ -1,0 +1,49 @@
+# The canonical observable declarations: energy, and its derivatives with
+# respect to the two inputs every model has.
+#
+# This file is the example every other observable copies. A new property is a
+# row here (or in a project's own file) and needs no code: the spec drives the
+# head, the loss term, the derivative names and the per-atom/per-graph padding.
+#
+# Units follow the project convention: eV and Å. The strain is dimensionless,
+# so the cell input carries "1".
+
+inputs:
+  # Atomic positions. A polar vector: it changes sign under inversion, which is
+  # what makes the energy gradient taken against it a 1o quantity as well.
+  - name: pos
+    irreps: "1o"
+    per_atom: true
+    units: "Å"
+
+  # The symmetric strain the cell derivative is actually taken against, not the
+  # nine cell entries. A symmetric rank-2 tensor is a scalar plus an l=2 part.
+  - name: cell
+    irreps: "0e+2e"
+    per_atom: false
+    units: "1"
+
+observables:
+  - name: energy
+    irreps: "0e"
+    per_atom: false
+    units: "eV"
+    # The std of the target, the successor of the legacy `std_scaling` entry.
+    # The legacy default was `rms_forces_scaling`, which scales the energy
+    # readout by the RMS of the *force* targets; that couples two observables
+    # through one number and cannot be written as a per-observable field. Which
+    # of the two v1 defaults to is a decision for the head, not for this file.
+    normalization: "std"
+    default_loss_weight: 1.0
+    derivatives:
+      # Named `forces`, reported as -dE/dpos.
+      - wrt: pos
+        units: "eV/Å"
+        normalization: "rms"
+        default_loss_weight: 100.0
+      # Named `stress`, reported as +dE/dstrain, divided by the cell volume by
+      # whatever computes it. The division is not part of the sign.
+      - wrt: cell
+        units: "eV/Å^3"
+        normalization: "none"
+        default_loss_weight: 1.0
--- a/packages/mace-core/src/mace_core/observables/spec.py
+++ b/packages/mace-core/src/mace_core/observables/spec.py
@@ -1,0 +1,320 @@
+"""Declaring an observable, an input, and the derivatives between them.
+
+Which quantities a legacy model computes is decided by a ladder of string
+comparisons on the model class name that sets six boolean flags, plus ten more
+``compute_*`` arguments on the forward signatures. Adding one property means
+editing several files. Here a property is a row: name it, say what shape it
+has, say whether there is one per atom or one per structure, and it is declared.
+
+Three objects, and they do different jobs:
+
+``InputSpec``
+    something the model is given -- positions, the cell, a magnetic moment, an
+    electronic temperature. Declared so that a derivative can be taken with
+    respect to it without new code.
+
+``ObservableSpec``
+    something the model produces and a loss can be written against.
+
+``DerivativeSpec``
+    derived, never declared: the result of asking an observable for its
+    derivative with respect to an input. Its name and sign come from
+    :mod:`mace_core.observables.derivatives`.
+
+``ObservableCatalogue`` holds a set of them and is what the defaults file loads
+into. Validation lives there rather than in the individual specs because the
+interesting errors are between rows: a derivative asked against an input nobody
+declared, or two rows whose derived names collide.
+"""
+
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+from mace_core.observables.derivatives import derivative_name, derivative_sign
+from mace_core.observables.grammar import (
+    IrrepTerm,
+    irreps_dimension,
+    parse_irreps,
+)
+
+__all__ = [
+    "NORMALIZATIONS",
+    "DerivativeRequest",
+    "DerivativeSpec",
+    "InputSpec",
+    "Normalization",
+    "ObservableCatalogue",
+    "ObservableSpec",
+]
+
+#: How a target is scaled before it reaches a head and a loss term. The
+#: per-observable successor of the legacy scaling registry, whose three entries
+#: were a global choice for the whole model. This package stores and validates
+#: the value; the head applies it in the output layer and the loss applies the
+#: matching term, both reading this one field, so there is never a second
+#: scaling mechanism to keep in step with it.
+Normalization = Literal["none", "std", "rms"]
+
+#: The accepted values, for error messages and for callers that enumerate them.
+NORMALIZATIONS: tuple[str, ...] = ("none", "std", "rms")
+
+_SCALAR = (IrrepTerm(multiplicity=1, degree=0, parity="e"),)
+
+
+def _check_name(value: str, kind: str) -> str:
+    if not value.isidentifier():
+        raise ValueError(
+            f"{kind} name {value!r} is not usable: a name must be a valid "
+            f"Python identifier, because it is also the key the value is "
+            f"stored under and part of any derivative name derived from it."
+        )
+    return value
+
+
+class InputSpec(BaseModel):
+    """Something the model is given, and can be differentiated against.
+
+    ``pos`` and ``cell`` are the two every model has. Anything else is declared
+    the same way, which is what makes ``d_energy_d_<feature>`` reachable for a
+    new feature without touching code.
+    """
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    name: str
+    irreps: str
+    #: ``True`` for one value per atom (positions, magnetic moments), ``False``
+    #: for one per structure (the cell, a total charge). This is what decides
+    #: whether a derivative taken against the input is padded per node or per
+    #: graph.
+    per_atom: bool
+    units: str = Field(min_length=1)
+
+    @model_validator(mode="after")
+    def _validate(self) -> InputSpec:
+        _check_name(self.name, "input")
+        parse_irreps(self.irreps, observable=self.name)
+        return self
+
+
+class DerivativeRequest(BaseModel):
+    """A derivative an observable asks for, and the loss settings it carries.
+
+    The name and the sign are not here: they are derived, and letting a
+    declaration override them would reintroduce the per-consumer naming this
+    abstraction removes. What a declaration does own is what a loss needs --
+    its weight, its normalization, and the unit string to report.
+    """
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    #: The name of the declared input to differentiate against.
+    wrt: str
+    default_loss_weight: float = Field(default=1.0, ge=0.0)
+    normalization: Normalization = "none"
+    #: Left to the declaration. Deriving it would mean unit algebra over the
+    #: quantity and the input, which this ticket does not own.
+    units: str | None = None
+
+    @model_validator(mode="before")
+    @classmethod
+    def _accept_bare_name(cls, value: object) -> object:
+        """``derivatives: [pos, cell]`` is the same as spelling out ``wrt``."""
+        if isinstance(value, str):
+            return {"wrt": value}
+        return value
+
+
+class DerivativeSpec(BaseModel):
+    """A derivative, as resolved by the catalogue. Never declared directly."""
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    name: str
+    #: The observable being differentiated.
+    of: str
+    #: The input it is differentiated against.
+    wrt: str
+    #: ``reported = sign * d(of)/d(wrt)``.
+    sign: int
+    #: Inherited from the input: a derivative against a per-atom input has one
+    #: value per atom, whatever the differentiated quantity is.
+    per_atom: bool
+    #: Known when the differentiated quantity is a single scalar, in which case
+    #: the gradient carries the input's irreps. ``None`` otherwise, because the
+    #: general case is a tensor product and the algebra is not this module's.
+    irreps: str | None
+    units: str | None
+    normalization: Normalization
+    default_loss_weight: float
+
+
+class ObservableSpec(BaseModel):
+    """One declared property: what it is, what shape it has, how it is scaled.
+
+    Any atomic or total spherical-tensor property declared here becomes
+    trainable with no new code: the spec drives the head, the loss term and the
+    padding classification.
+    """
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    name: str
+    #: The spherical-tensor shape, in the grammar of
+    #: :mod:`mace_core.observables.grammar`.
+    irreps: str
+    #: ``True`` for one value per atom, ``False`` for one per structure. Every
+    #: output carries this, which is what lets padding be added and removed
+    #: without a consumer keeping its own list of which names are which.
+    per_atom: bool
+    #: Project convention: eV, Å.
+    units: str = Field(min_length=1)
+    normalization: Normalization
+    default_loss_weight: float = Field(default=1.0, ge=0.0)
+    #: The derivatives this observable asks for. Naming works for any declared
+    #: input whether or not it is listed here; listing it is what says the
+    #: model should compute it.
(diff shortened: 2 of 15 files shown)


packages/mace-core/src/mace_core/defaults/observables.yaml
```diff
@@ -0,0 +1,49 @@
...
+    default_loss_weight: 1.0
+    derivatives:
+      # Named `forces`, reported as -dE/dpos.
+      - wrt: pos
+        units: "eV/Å"
+        normalization: "rms"
+        default_loss_weight: 100.0
+      # Named `stress`, reported as +dE/dstrain, divided by the cell volume by
+      # whatever computes it. The division is not part of the sign.
+      - wrt: cell
```
steffen-wedig: The stress should really be defined as derivative wrt strain, not the cell vectors, right? This might make the config more tricky, because the zero strain in get symmetric displacement is not an input to the model.

packages/mace-core/src/mace_core/observables/spec.py
```diff
@@ -0,0 +1,320 @@
...
+    #: Inherited from the input: a derivative against a per-atom input has one
+    #: value per atom, whatever the differentiated quantity is.
+    per_atom: bool
+    #: Known when the differentiated quantity is a single scalar, in which case
+    #: the gradient carries the input's irreps. ``None`` otherwise, because the
+    #: general case is a tensor product and the algebra is not this module's.
+    irreps: str | None
+    units: str | None
+    normalization: Normalization
+    default_loss_weight: float
```
steffen-wedig, line 158: From my view both the loss_weight and the normalization are things that should be in the training_config/ the loss config. The default loss weight should just be a default values on a config field maybe.

aacostadiaz: Agreed on the weights. They can change between stages so they belong in the loss config.

I’d move normalization to the model config. If we scale the energy head forces pick up that scale, so configuring a separate force scale wouldn’t make sense. We need to configure how we get the scale for each head, including from another target (rms_forces_scaling already uses force RMS to set the energy scale)

What do you think? That leaves ObservableSpec with name, irreps, per_atom, and units

steffen-wedig: Yes, that sounds good to me. The normalization values should move as buffers with the models, so I agree that model config should be the correct place for the the normalization to go aswell.

PR #1726, opened 2026-09-13
Aperiodic neighbour lists must not depend on where the molecule sits

lollcat:
`get_neighborhood` is not translation invariant for non-periodic systems. The same
molecule placed away from the coordinate origin comes back with edge vectors many
times the cutoff, at an unchanged edge count.

## Reproduction — the neighbour list

```python
import numpy as np
from mace.data.neighborhood import get_neighborhood

atoms, cutoff = molecule("H2O"), 5.0
for offset in (0.0, 100.0):
    positions = atoms.get_positions() + offset
    edge_index, shifts, _, _ = get_neighborhood(positions, cutoff, pbc=(False, False, False))
    d = positions[edge_index[1]] - positions[edge_index[0]] + shifts  # D = r_j - r_i + S.cell
    print(f"offset {offset:6.1f} A   edges {edge_index.shape[1]}   max |D| {np.linalg.norm(d, axis=1).max():8.3f} A")
```

On `develop`:

```
offset    0.0 A   edges 6   max |D|    1.526 A
offset  100.0 A   edges 6   max |D|  155.914 A
```

With this PR both rows read `1.526 A`.

## Reproduction — what it does to a calculator

```python
import numpy as np
from mace.calculators import mace_off

calc = mace_off(model="small", default_dtype="float64", device="cpu")  # r_max = 4.5 A

for offset in (0.0, -6.0, 100.0):
    atoms = molecule("H2O")
    atoms.positions += offset          # a rigid translation of the same molecule
    atoms.calc = calc
    print(f"offset {offset:6.1f} A   E {atoms.get_potential_energy():14.6f} eV   "
          f"max|F| {np.abs(atoms.get_forces()).max():9.6f} eV/A")
```

On `develop`:

```
offset    0.0 A   E   -2081.116129 eV   max|F|  0.651139 eV/A
offset   -6.0 A   E   -2071.077623 eV   max|F|  0.000000 eV/A
offset  100.0 A   E   -2071.077623 eV   max|F|  0.000000 eV/A
```

With this PR all three rows read `-2081.116129 eV` and `0.651139 eV/A`, agreeing to
2e-13 eV/A.

The displaced molecule comes out **10.04 eV (231 kcal/mol) higher, with forces exactly
zero**: every edge is longer than the cutoff, so the model sees three non-interacting
atoms and returns the sum of their atomic references. Note the offset that does it —
`-6.0`, not `100.0`.

## How far from the origin is far enough

The box for a non-periodic axis spans `[0, extent + 2 * cutoff + 1]`, and matscipy
tolerates roughly half a box-width of overhang before the shifts it reports stop
matching. The threshold therefore scales with the cutoff and with the molecule's own
extent — nothing the caller controls or can see:

| cutoff | per-axis boxes for H2O | first offset that corrupts it |
|---|---|---|
| 3.0 A | 7.0 / 8.5 / 7.6 A | −3.35 A |
| 5.0 A | 11.0 / 12.5 / 11.6 A | −5.35 A |
| 8.0 A | 17.0 / 18.5 / 17.6 A | −8.35 A |

A structure written out in the positive octant, a solvent box recentred on its solute,
or a conformer set carrying absolute coordinates can all be outside it.

## Which code paths this reaches

`get_neighborhood` has exactly one caller, `AtomicData.from_config`, so the exposure is
everything that builds a graph in Python:

- **Affected** — `MACECalculator` and `MagneticMACECalculator`, and hence every
  foundation wrapper that returns one: `mace_mp`, `mace_off`, `mace_omol`,
  `mace_polar`, `mace_anicc`, `mace_mdp`.
- **Affected** — training and evaluation, not just inference: `run_train.py`,
  `eval_configs.py`, and the HDF5/LMDB dataset builders all go through
  `from_config`, so a non-periodic training configuration sitting off the origin is
  corrupted at the point its graph is built.
- **Not affected** — `LAMMPS_MACE` and the ML-IAP wrapper, which are handed the
  neighbour list by LAMMPS, and `MACETorchSimModel`, which builds its own via
  `torchsim_nl`.
- **Not affected** — fully periodic systems. Only non-periodic axes get a fictitious
  box, so bulk crystals are untouched; molecules in vacuum, and the vacuum axis of
  slabs, are the exposure.

## Cause

For a non-periodic axis, `get_neighborhood` builds a fictitious box so matscipy has
something to bin into. #1533 sized that box from the atom extent
(`extent + 2 * cutoff + 1`) rather than from `max(|positions|) * 5 * cutoff`, which was
correct for its purpose — the old formula produced enormous cells for molecules far
from the origin and blew up k-space electrostatics.

But it fixed the box's **size** without changing its **placement**: the box is still
anchored at the origin, and the discarded formula's dependence on absolute coordinates
was precisely what had kept a displaced molecule inside it. matscipy wraps whatever
lies outside the box and reports the wrap in `unit_shifts`. Those shifts are added back
to the caller's *unwrapped* positions downstream (`D = r_j - r_i + S.cell`), so a 1.5 A
bond becomes a 156 A vector.

Wrapping is a bijection on the atoms, so the edge **count** is unchanged. Nothing that
checks edge counts notices, and no existing test displaces a molecule far enough to
leave its own box.

## Fix

Translate the positions handed to the search so the non-periodic axes start at 0. The
offset is internal to the search and rigid, so it leaves every interatomic distance —
and hence the neighbour list — alone, and the caller's positions are untouched.
Periodic axes keep their own coordinates, where wrapping is meaningful and the shifts
are real.

## Note on `mace-reforge`

`mace/data/neighborhood.py` on `mace-reforge` is byte-identical to the version this
fixes. Since the frozen `mace/` tree there is the reference the v1 implementation is
measured against, that reference currently encodes this behaviour for non-periodic
configurations away from the origin. Flagging it rather than touching that branch.

## Test

Two, beside the existing `test_nonperiodic_cell_is_extent_based`, which pins the box's
size but not its edges.

- `test_aperiodic_edges_are_translation_invariant` asserts the neighbour list is
  unchanged by a rigid translation, that no shift is non-zero when nothing is periodic,
  and that `|D|` respects the cutoff.
- `test_partial_pbc_is_translation_invariant_along_the_vacuum_axis` pins the other half
  of the fix — that periodic axes keep their own coordinates and their real wrap-around
  shifts — by displacing a slab along its vacuum axis.

Both use the existing `tests/neighbour_oracle.py` helpers. Without the fix they fail in
three of five parametrised cases; with it `tests/unit` passes (900 passed, 22 skipped),
including the brute-force oracle comparisons on non-orthogonal partially-periodic cells.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aacostadiaz (approved):
Thanks for fixing this

Issue #1556, opened 2026-08-04
CORE-2 — Config system skeleton + model metadata schema


aacostadiaz:
**Depends on:** INF-1 (#1550) · **Blocks:** CFG-1 (#1574), CLI-1 (#1579), FM-3 (#1588)

**Context:** Today the entire training configuration surface is one flat argparse namespace: `mace/tools/arg_parser.py` is 1417 lines declaring **210 argparse dests across its two parsers, counted per parser** — 184 on `build_default_arg_parser` (`:14`) and 26 on `build_preprocess_arg_parser` (`:1239`), 188 of them distinct since 22 are declared in both — carried by 220 option strings, each count including the configargparse `--config` file flag (`:22-27`, `:1247-1252`, registered with `parser.add` rather than `add_argument`). A YAML file **is** a first-class input path — `configargparse` is a hard install dependency (`setup.cfg:32`), not an optional one — but there is no typed validation: bad keys and bad values surface deep inside the run, and what a model was actually trained with is not recorded in the model itself. The closest legacy gets is the TorchScript side artifact, which embeds `commit.txt` and a reflected `config.yaml` as `_extra_files` (`mace/cli/run_train.py:1184-1189`); the plain `.model` checkpoint that everyone actually loads is a bare pickle of the module (`:1183`) with no config, no provenance and no schema version. v1 rebuilds configuration around **Pydantic + pydantic-settings**: TOML/YAML/JSON input, **config-file-first** with dotted CLI overrides, unknown keys as hard errors, and the **fully resolved config (all defaults filled) saved into model metadata — mandatory for ALL models**, not just foundation models (decided; nearly free once this machinery and CLI-1 (#1579) exist). This ticket builds the *machinery* — the base config class and the metadata schema — not the full training schema: CFG-1 (#1574) later defines the real `data/model/loss/optimizer/schedule/runtime/finetune` sections on top of it, CLI-1 (#1579) wires the hierarchical `mace train/eval/model/data/export` CLI (including exporting the stored resolved config back to YAML and automatic citation printing), and FM-3 (#1588)'s model registry consumes `ModelMetadata` (DOIs, recommended/deprecated labels, query API). Skim `mace/tools/arg_parser.py` only for scale — do **not** port it; the legacy flag surface is mapped later (CFG-1 (#1574)/CLI-1 (#1579)).

**Interface & constraints:**

1. **`mace_core.config.ReforgeBaseConfig`** (pydantic-settings base class all v1 configs subclass):
[...]

PR #1733, opened 2026-09-17
core-2

arnon-1:
Hi,

Fixes https://github.com/ACEsuit/mace/issues/1556

Adds `mace_core.config.ReforgeBaseConfig` and `mace_core.metadata.ModelMetadata`.

Deviations from the ticket:
- No pydantic-settings as it doesn't entirely cover the use-case and writing a wrapper ends up in more convoluted code than just writing a simple parser.

- test files use the package prefix e.g. `test_mace_core_config.py` so not literally the same as the verify command in the ticket but adheres to the packages README rule.

Some choices not mentioned in the ticket (non-exhaustive).
- Some edge-cases can still raise errors different from ConfigError or MetadataSchemaError, but it doesn't make sense to handle every single edge case explicitly as the code would become quite large.

- I am rejecting a number of patterns that results in non-deterministic json (like sets) or are bug-prone (like overloading parameters so that pydantic implicitly selects classes).

- Field groups adapted a bit to take into account multi-dataset multi-head models. Adds a bit of overhead for single-head model data. Can change it if someone disagrees with the way I did it here.

- Also added a parent model field for fine-tuning or distillation (which uses ModelMetadata again recursively), I thought this was nice for provenance, but can remove it if seen as unneeded.

---

> [!NOTE]
> **Medium Risk**
> New foundational config and checkpoint-metadata contracts will shape all future v1 training and loading; behavior is well-tested but downstream adopters must match precedence, kinds syntax, and metadata schema version 1.
> 
> **Overview**
> Replaces the **mace-core** scaffold with the v1 **configuration** and **model metadata** contracts (issue #1556).
> 
> **`ReforgeBaseConfig` / `ConfigSection`** add a custom Pydantic loading path: one TOML/YAML/JSON file plus `--dotted.cli` overrides (defaults → file → CLI, no env vars). Unknown keys raise `ConfigError` with typo suggestions; discriminated “kinds” sections use kind-as-key in files/CLI with merge rules and `ConfigWarning` when a later layer switches variant. Schemas are guarded at class definition (no sets, aliases, undiscriminated section unions). Exports include `to_resolved_dict()` / `to_user_dict()` for JSON-stable, round-trippable configs.
> 
> **`ModelMetadata`** defines schema version 1 checkpoint sidecar data: user vs resolved config (`ConfigRecord.from_config`), provenance, per-source and per-head summaries (including E0s), citations, optional recursive **`parents`** for fine-tune/distill lineage, strict `extra=forbid`, and `to_json()` / `from_json()` with version mismatch and lossy-value rejection. **`format_citations()`** renders a numbered bibliography block.
> 
> Package **dependencies** are now `pydantic`, `pyyaml`, and `tomli`; the top-level **`mace_core`** package re-exports the public types. Large pytest suites cover file formats, CLI/kinds behavior, and metadata round-trips (including no torch/jax import leak).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5339d4f3a47f7840b087514b56b764857fc3678. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/mace-core/src/mace_core/config/base.py
```diff
@@ -0,0 +1,368 @@
...
+                name,
+                "is typed as a set; set order is not stable across runs. Use a list",
+            )
+        sections = {section for section, _ in _sections_in(field.annotation)}
+        if len(sections) > 1:
+            reject(
+                name,
+                "is a union of sections; give each alternative its own optional "
+                "field, e.g. `huber: HuberLoss | None`",
+            )
```
steffen-wedig, line 131: This rejects Annotated Unions of pydantic models. If every alternative of any config has its own optional field, these configurations will become extremely wide. Pydantic with Annotated Unions works extremely well, and I think it would be an absolute waste not to use them.  https://gist.github.com/steffen-wedig/2356d642f052c7f2c53c6f3407ac2b9a

packages/mace-core/src/mace_core/config/base.py
```diff
@@ -0,0 +1,368 @@
...
+            unknown.append(_unknown_key_message(name, valid))
+            continue
+
+        if value == "null" or value.startswith(("[", "{")):
+            try:
+                value = json.loads(value)
+            except ValueError as error:
+                raise ConfigError(
+                    f"override --{name} is not valid JSON: {error}"
+                ) from error
```
steffen-wedig, line 296: It seems to me like all of this is cli functionality that should go into the cli module, not the config - this all just parses the arguments in the command. If the cli module already returns a dict with overides, then config/base.py only needs to apply the update with _deep_update()

packages/mace-core/src/mace_core/config/base.py
arnon-1: So in general, I tend to be not very enthusiastic about allowing multiple ways for doing the same things. But I guess this does have advantages.
So I think I would make the discriminator field non-optional as otherwise it allows for unpredictable behaviour as far as I am aware.
Unless you think unions without discriminator would be useful?

steffen-wedig: Yes, the disciminator field should be mandatory. There are lots of parts of the code where there are multiple ways of doing things. e.g. different loss functions, different radial functions, different interaction blocks, etc, right? But you will set one of them in your config. If you give each alternative its own field, whats stopping me from setting the huber loss field, and the universal loss field at the same time? the LossConfig field with a discriminated union prevents that, imo. But also happy to discuss further.

bernstei: In principle, you could implement something equivalent to argparse's `mutually_exclusive_group`, which looks like separate fields.  But I also prefer a union when the structure should be in some limited sense content agnostic. I suppose you also could do (perhaps the equivalent of) defining a superclass, and make the specific possibilities derived from that.

packages/mace-core/src/mace_core/config/base.py
arnon-1: I would kind of agree with you on this. But the cli seems specific for jax and torch right now. the mace-core doesn't have a cli module. It wouldn't make sense to duplicate this function for torch and jax. So the cli passes arguments related to config to the config module.

packages/mace-core/src/mace_core/config/base.py
arnon-1: Yes sure, I guess the advantages indeed outweigh the disadvantages here.

packages/mace-core/src/mace_core/config/base.py
```diff
@@ -0,0 +1,809 @@
...
+class ConfigWarning(UserWarning):
+    """A section of one kind is ignored because a later layer selected another
+    kind. `warnings.simplefilter("error", ConfigWarning)` makes it an error."""
+
+
+class ConfigSection(BaseModel):
+    """A nested section of a configuration: a table in the file, a dotted
+    prefix on the command line. Unknown keys are errors here too."""
+
+    model_config = ConfigDict(extra="forbid")
```
aacostadiaz: `inf` and `nan` come out as `null` which would put the wrong value into the model metadata

```suggestion
    model_config = ConfigDict(extra="forbid", ser_json_inf_nan="constants")
```

packages/mace-core/src/mace_core/config/base.py
```diff
@@ -0,0 +1,809 @@
...
+    return any(_contains_a_set(arg) for arg in get_args(annotation))
+
+
+_NONE_IS_A_KIND = (
+    "default to a variant, or for none of the kinds add an empty variant, "
+    'kind: Literal["none"], and default to that'
+)
+
+
+def _check_schema(model: type[BaseModel]) -> None:
```
aacostadiaz: I think these checks run before pydantic resolves forward references, so a forward-ref field gets an error that contradicts itself:

    class CA(ConfigSection):
        q: "CB | None" = None
    -> TypeError: CA.q defaults to None but its type does not admit None; add | None

The other half is the one that worries me. A forward-ref field with a non-None default slips past every guard and then a plain BaseModel behind it silently drops unknown keys:

    class Root(ConfigSection):
        leaky: "Leaky" = Field(default_factory=lambda: Leaky())

    class Leaky(BaseModel):      # not a ConfigSection, so it allows extra keys
        a: int = 1

    Root.model_rebuild()

    class Demo(ReforgeBaseConfig):
        root: Root = Field(default_factory=Root)

    # demo.json is {"root": {"leaky": {"a": 2, "TYPO": 9}}}
    print(Demo.load("demo.json", []).to_resolved_dict())
    # -> {'root': {'leaky': {'a': 2}}}.    # TYPO gone

With the class defined before `Root`, the same shape is caught properly, so it's the unresolved ForwardRef that opens the hole. Running the checks at `model_rebuild()` time or re-checking once the ForwardRef resolves would cover both halves.

aacostadiaz:
@cursor review verbose=true

packages/mace-core/src/mace_core/config/base.py
steffen-wedig: I dont quite understand why the cli is specific to jax and torch - the backend selection between torch and jax happens downstream of configuring models, right?

arnon-1: So I believe the current plan is that the core part decides the syntax for the cli arguments (or at least the syntax for building the cli syntax). But the cli is specific to torch and jax. The reason I believe is that torch also contains training apart from inference. So in the torch cli, the training part contains the most options and would not be shareable with the jax part. Though the inference part I think is in principle shareable, though I believe this is not what the current plan describes. Not sure in how far this has already been considered. Though I think the jax scope might still shrink for the initial phase?

packages/mace-core/src/mace_core/config/base.py
```diff
@@ -0,0 +1,621 @@
...
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs: Any) -> None:
+        super().__pydantic_init_subclass__(**kwargs)
+        check_model_config(cls)
+        if cls.__pydantic_complete__:
+            check_fields(cls)
+
+    @model_validator(mode="before")
+    @classmethod
+    def _kinds_fields_to_tagged_form(cls, data: Any) -> Any:
```
aacostadiaz: Could we use pydantic's standard tagged form for kinds, loss: {kind: huber, delta: 0.1}, instead of the kind name as a key? It would remove the conversion in both directions, its errors and its warnings, and most of the kinds tests. We lose keeping the settings of unused kinds in the same file, but I am not sure we need that.

arnon-1:
Ok so as requested, I dropped most of the requirements in the original ticket. So the current cli is pretty basic but the config itself is basically the same. I do still need to review this more carefully.

